PR #10, opened 2026-02-26
Fix bearer token validation to accept both admin_token and webhook_secret

mbradley:
## Summary

- Bearer token validation used an `or_else` fallback chain that only checked `webhook_secret` when `admin_token` was missing from the secret store
- Since `admin_token` exists, the moderation-service (which authenticates with `webhook_secret`) always got 401 on the admin content bypass endpoint
- Now checks both tokens independently so either credential works

Required for divine-moderation-service PR #10 (admin video proxy for blocked content).

## Test plan

- [x] Deployed to Fastly production (version 209)
- [x] Moderation review queue now displays blocked videos
- [ ] Verify admin dashboard login still works with admin_token

### DIFF
--- a/src/admin.rs
+++ b/src/admin.rs
@@ -50,51 +50,41 @@ pub fn validate_admin_auth(req: &Request) -> Result<()> {
 }
 
 /// Validate Bearer token from Authorization header
+/// Accepts either admin_token or webhook_secret from the Fastly Secret Store.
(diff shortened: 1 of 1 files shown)


rabble:
Closing in favor of PR #12 which addresses the same bug reintroduced by commit 1cb3ee8.

PR #24, opened 2026-03-12
feat: add c2patool trust checking to process-blog cloud run

n8fr8:
New files:
  - cloud-functions/process-blob/Dockerfile — Container image with c2patool v0.26.33 binary, Python 3.12, gunicorn/Flask
  - cloud-functions/process-blob/trust_anchors.pem — Placeholder for trusted C2PA signer CA certificates (replace with actual ProofSign/Guardian Project certs)

  Modified files:
  - cloud-functions/process-blob/main.py — Complete rewrite with:
    - Flask app for Cloud Run (Eventarc HTTP POST), preserving Cloud Function entry point
    - check_c2pa_trust() — Downloads blob to temp file, runs c2patool --detailed to read manifest, then c2patool trust --trust_anchors to validate trust chain
    - C2PA runs before SafeSearch to save Vision API costs on untrusted content
    - Three modes: C2PA_MODE=off (default, no checking), log (check + record, never reject), enforce (reject untrusted → status restricted)
    - C2PA results included in metadata webhook payload
  - cloud-functions/process-blob/requirements.txt — Added flask and gunicorn
  - scripts/deploy-cloud-function.sh — Converted from gcloud functions deploy to gcloud run deploy with Eventarc trigger

  To activate:
  1. Add your trusted signer certificates to trust_anchors.pem
  2. Set C2PA_MODE=log for monitoring, then C2PA_MODE=enforce when ready to reject untrusted uploads
  3. Deploy with scripts/deploy-cloud-function.sh


cloud-functions/process-blob/main.py
```diff
@@ -26,31 +57,287 @@ def process_blob(event, context):
...
+        print("c2patool binary not found")
+        return None
+
+
+def _run_c2patool_trust(file_path):
+    """Run c2patool trust check against configured trust anchors. Returns True if trusted."""
+    try:
+        proc = subprocess.run(
+            ['c2patool', '--detailed', file_path, 'trust',
+             '--allowed_list', C2PA_TRUST_ANCHORS],
```
rabble: The implementation and docs still describe C2PA_TRUST_ANCHORS / trust_anchors.pem as a CA trust-anchor bundle, but this invocation is using --allowed_list, which is a different trust model. That turns the PEM into a pinset of explicitly allowed signer certificates, so signer rotation under the same ProofSign CA will start failing verification even though the chain is still valid.\n\nIf the intent is CA-based verification, this should use the trust-anchor option consistently and the PEM should contain only the CA material. If the intent is explicit signer allowlisting, then the env var, file name, and README should all be renamed because they currently describe the wrong security model.

rabble (changes requested):
I pulled the branch locally and went through the new Cloud Run / C2PA path end to end. I think this needs changes before merge for three production issues:\n\n1. The Eventarc trigger is bucket-wide, but the new derivative skip logic only matches thumbnails/ and top-level hls/. Blossom's actual generated objects are {hash}.jpg, {hash}/hls/*, and {hash}/vtt/*, so generated artifacts will still re-enter this service.\n2. The trust verification path is treating the configured PEM as an allowlist of signer certs rather than as CA trust anchors. That will make signer rotation fail even when the signer chains to the same ProofSign CA.\n3. The default C2PA max size is 2 GiB while the deploy script gives the service 1 GiB of memory. Files in that gap can still be downloaded to temp and kill the instance.\n\nDetails are inline with suggested fix directions.

rabble:
Closing in favor of #37 which includes all commits from this PR plus improved derivative filtering.

The `_is_generated_derivative()` regex in #37 correctly matches all HLS segments (`{hash}/hls/*`) and VTT files (`{hash}/vtt/*`), not just `master.m3u8` and `main.vtt`.

@n8fr8 — great work on the C2PA integration. The remaining blocker for #37 is the operational rollout plan (Cloud Function → Cloud Run migration). Can you confirm the deployment path per the questions in #37?

PR #44, opened 2026-03-25
fix: return 403 ContentRestricted for moderated blobs instead of 404

### LINKED ISSUE #41: fix: return 403 for moderation-restricted content instead of 404
## Problem

All serving endpoints return `404 {"error":"Blob not found"}` for both genuinely missing content AND moderation-restricted/banned content. The mobile app can't distinguish between them and shows "Failed to load video" for restricted content instead of a meaningful moderation message.

## Evidence

- 6% of sampled videos are `restricted` or `banned` by the AI moderation system
- The app already handles `401` differently (shows "Age-restricted" + "Verify Age" button)
- PR #33 adds `blocks_public_access()` / `requires_owner_auth()` helpers that are the right hook point
- `VideoModerationStatusService` exists on the mobile side but is only wired to badge UI

## Proposed fix

Return `403 {"error":"Content restricted","status":"restricted"}` instead of `404 {"error":"Blob not found"}` when metadata exists and status is `restricted` or `banned`. Keep `404` for genuinely missing blobs.

The mobile app can then add `_is403Error()` following the existing `_is401Error()` pattern.

## Related

- PR #33 (soft-deleted blob access control)
- divine-mobile #2269 (Content Moderation epic)

dcadenas:
## Summary
- Add `ContentRestricted` error variant for moderated content
- Return 403 for moderated blobs instead of generic 404

Closes #41

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -203,16 +203,16 @@ fn handle_get_blob(req: Request, path: &str) -> Result<Response> {
         if let Ok(Some(meta)) = get_blob_metadata(video_hash) {
             if !is_admin {
                 if meta.status == BlobStatus::Banned {
-                    return Err(BlossomError::NotFound("Blob not found".into()));
+                    return Err(BlossomError::ContentRestricted("banned".into()));
                 }
                 if meta.status == BlobStatus::Restricted {
                     // Check if requester is owner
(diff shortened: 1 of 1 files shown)


dcadenas:
Premature — closing. Will re-create after proper Forgejo review.

PR #46, opened 2026-03-25
fix: cloud-run-transcoder misconfigurations

### LINKED ISSUE #32: fix: cloud-run-transcoder infrastructure misconfigurations
## Summary

The production transcoder (`divine-transcoder` Cloud Run service) has several infrastructure misconfigurations that were discovered while investigating the `-bf 0` fix in #31. These issues affect deployment reliability, performance, and operational safety.

**Update (2026-03-17):** The immediate B-frame fix from #31 has been merged and deployed (revision `divine-transcoder-00024-htj`). Production logs confirm `-bf:v:0 "0"` and `-bf:v:1 "0"` in ffmpeg commands — CPU transcodes now produce HLS-compatible output. The remaining items below are infrastructure concerns that still need to be addressed.

### 1. Transcoder deployed in wrong GCP project

The live `divine-transcoder` service runs in the `poc` project (`rich-compiler-479518-d2`, project number `149672065768`) instead of `dv-platform-prod`. The Fastly edge server has this URL hardcoded in `src/main.rs:1593`.

### 2. No GPU hardware — blocked on quota

Despite the codebase expecting GPU acceleration, the live service has no GPU. Logs on every startup show:

```
WARNING: The NVIDIA Driver was not detected. GPU functionality will not be available.
GPU acceleration: disabled (CPU fallback)
```

All transcodes run on CPU via `libx264`. The `deploy.sh` script references `--gpu 1 --gpu-type nvidia-l4` but the service was deployed separately without GPU by `rabble@divine.video`.

Attempting to enable GPU via `gcloud run services update --gpu 1 --gpu-type nvidia-l4` fails because the project lacks Cloud Run-specific GPU quota (separate from Compute Engine GPU quota). Error: _"You do not have quota for using GPUs without zonal redundancy."_ GPU quota must be requested at g.co/cloudrun/gpu-quota before GPU can be enabled.

### 3. `USE_GPU` env var never set

Even if GPU hardware were attached, `USE_GPU=true` is not in the service's environment variables, so the code would default to CPU anyway.

### 4. `deploy.sh` is out of sync with live service

Key differences between `deploy.sh` and the actual running config:

| Setting | `deploy.sh` | Live service |
|---------|-------------|-------------|
| Image registry | `gcr.io` | `us-central1-docker.pkg.dev` |
| Memory | 16Gi | 8Gi |
| Concurrency | 8 | 320 |
| Transcription model | `whisper-1` | `gpt-4o-mini-transcribe` |
| Retry env vars | set | not set |
| min-instances | 1 | not set (default 0) |

Additionally, `deploy.sh` uses `--set-env-vars` which **replaces all env vars**, making it unsafe to re-run without first reconciling with the live configuration.

### 5. No Infrastructure-as-Code configuration

The `divine-transcoder` Cloud Run service has no IaC configuration. It is not tracked in the `divine-iac-coreconfig` repo at all — the only related reference is a docs page at `docs/how-to/services/divine-blossom.md`, which doesn't mention the transcoder. The service was set up manually by Rabble via `gcloud` CLI. This should be brought under IaC (Terragrunt or at minimum a tracked config) as part of the infrastructure cleanup.

## Impact

- All transcodes run on CPU, which is significantly slower than GPU
- GPU enablement is blocked on a quota request to Google (g.co/cloudrun/gpu-quota)
- `deploy.sh` cannot be safely re-run without risk of breaking the live service
- Service is in a non-production GCP project with potentially different IAM, billing, and monitoring
- Infrastructure state is not reproducible from the repo
- No IaC means infrastructure drift is invisible and changes are not auditable

## Context

Discovered during investigation for the `-bf 0` B-frame fix (#31). That fix addressed the immediate symptom (CPU transcoding producing B-frames incompatible with HLS) and is now deployed and confirmed working. The root cause — that the transcoder should be running on GPU — and the broader infrastructure concerns remain open.

dcadenas:
## Summary
- Centralize Cloud Run host constants via config store
- Make deploy.sh safe to re-run and add USE_GPU env var
- Parameterize hardcoded GCP project in utility scripts

Closes #32

### DIFF
--- a/cloud-run-transcoder/backfill-vtt-direct.sh
+++ b/cloud-run-transcoder/backfill-vtt-direct.sh
@@ -11,9 +11,10 @@ RESET_STALE="${RESET_STALE:-false}"  # set to true to re-trigger stuck "processi
 MAX_RETRIES=3
 
 # Auth: prefer webhook secret from GCP, fall back to env var or cookie
+GCP_PROJECT="${GCP_PROJECT:-rich-compiler-479518-d2}"
 if [[ -z "$WEBHOOK_SECRET" ]]; then
-    echo "Fetching webhook secret from GCP..."
-    WEBHOOK_SECRET=$(gcloud secrets versions access latest --secret=webhook_secret --project=rich-compiler-479518-d2 2>/dev/null || true)
+    echo "Fetching webhook secret from GCP (project=${GCP_PROJECT})..."
+    WEBHOOK_SECRET=$(gcloud secrets versions access latest --secret=webhook_secret --project="${GCP_PROJECT}" 2>/dev/null || true)
 fi
 
 if [[ -n "$WEBHOOK_SECRET" ]]; then
--- a/cloud-run-transcoder/backfill-vtt-force.sh
+++ b/cloud-run-transcoder/backfill-vtt-force.sh
@@ -8,12 +8,13 @@
 #
 # To resume from a specific offset: OFFSET=500 ./backfill-vtt-force.sh
 
-URL="https://separately-robust-roughy.edgecompute.app/admin/api/backfill-vtt"
+URL="${BACKFILL_URL:-https://separately-robust-roughy.edgecompute.app/admin/api/backfill-vtt}"
+GCP_PROJECT="${GCP_PROJECT:-rich-compiler-479518-d2}"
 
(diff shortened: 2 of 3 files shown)


dcadenas:
Premature — closing. Will re-create after proper Forgejo review.

PR #75, opened 2026-04-09
fix: add orientation backfill script for pre-Feb-7 portrait videos

dcadenas:
## Summary

- Add `scripts/backfill-orientation.sh` to detect and fix portrait videos that were transcoded as landscape before the orientation-aware encoding fix (1d49c0b, Feb 7)
- Add state file patterns to `.gitignore`

## Context

The old transcoder used hardcoded `scale_cuda=1280:720` / `scale=1280:720` for all videos regardless of orientation. Portrait videos (coded height > width) were forced into landscape 720p/480p variants. The orientation fix (1d49c0b) landed Feb 7, but existing transcodes were not retroactively corrected. The March 19 MP4 backfill (PR #36) remuxed the already-wrong .ts files with `-c:v copy`, carrying the bad orientation into the progressive MP4 variants that the mobile app now uses by default.

## What was done

- Scanned all 292,411 videos in the GCS bucket
- Found 155 natively portrait videos with landscape 720p variants
- All affected .ts files were created before Feb 7 (confirmed by Last-Modified timestamps)
- Re-transcoded all 155 by deleting old HLS from GCS and calling the transcoder, which now produces correct orientation
- Purged CDN cache via `curl -X PURGE` on specific URLs + `fastly service purge --all`
- Videos with rotation metadata (coded landscape + Display Matrix rotation) were excluded — the old GPU transcoder handled those correctly since `scale_cuda` operated on unrotated frames

## The script

Idempotent detection + deletion tool following the `backfill-fmp4.sh` pattern:
- `--dry-run`: scan and list affected hashes without deleting
- `--hash <sha256>`: check/fix a single video
- Batch mode: streams hashes from GCS, skips already-checked via awk hash set, concurrent probing

Re-transcoding and CDN purging were done separately (Cloud Run transcoder + Fastly purge). The script handles the detection and HLS deletion step.

## Closes

divinevideo/divine-mobile#2858

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -54,6 +54,9 @@ migration_progress.json
 scripts/backfill-fmp4-all-hashes.txt
 scripts/backfill-fmp4-done.txt
 scripts/backfill-fmp4-failed.txt
+scripts/backfill-orientation-done.txt
+scripts/backfill-orientation-affected.txt
+scripts/backfill-orientation-failed.txt
 
 # Local/ephemeral
 .claude/
--- a/scripts/backfill-orientation.sh
+++ b/scripts/backfill-orientation.sh
@@ -1,0 +1,211 @@
+#!/bin/bash
+# ABOUTME: Fix portrait videos that were transcoded as landscape before the orientation fix (1d49c0b, Feb 7 2026).
+# ABOUTME: Detects orientation mismatch between original blob and 720p variant, deletes bad HLS so on-demand re-transcoding rebuilds correctly.
+# ABOUTME: Idempotent: tracks done hashes, skips on re-run.
+#
+# Prerequisites:
+#   gcloud config configurations activate divine  # need GCS access to divine-blossom-media bucket
+#   ffprobe installed
+#
+# Usage:
+#   ./scripts/backfill-orientation.sh --dry-run     # detect and list affected hashes without deleting
+#   ./scripts/backfill-orientation.sh               # detect and delete bad HLS for affected hashes
+#   ./scripts/backfill-orientation.sh --hash abc123  # check/fix a single hash
+#
+# State files (in scripts/):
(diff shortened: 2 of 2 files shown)


rabble:
## Review

Overall looks good — clean, idempotent backfill script following existing patterns. A few notes:

### 1. PROBE_FAIL hashes get marked done (line 143)

After a `PROBE_FAIL`, the hash goes to `$FAILED_FILE` but still falls through to `echo "$hash" >> "$DONE_FILE"`. This means a hash that failed probing (temporary network issue) will be skipped on re-run. If the intent is "retry failed on re-run", move the `$DONE_FILE` append into each case arm rather than after the case block.

### 2. `gcloud config configurations activate divine` is side-effectful

Line 176 switches the active gcloud config globally. If the script crashes before line 193 restores it, the user's gcloud config stays switched. Consider using `--project` flags on each command instead (like `delete_hls` already does), or adding a `trap 'gcloud config configurations activate default 2>/dev/null' EXIT`.

### 3. `total` counter lost in subshell

The `while read` pipeline creates a subshell, so `total` incremented inside won't reflect in the parent after the loop. The file-based counts (`$DONE_FILE`, `$AFFECTED_FILE`) still work fine — this is cosmetic only.

### 4. Good to merge

The backfill is already done (155 videos fixed per the description), so this is tooling for posterity. The script is well-structured with dry-run, single-hash mode, and progress reporting. LGTM with the minor notes above.

rabble (approved):
LGTM — well-scoped one-off backfill that follows the established backfill-fmp4.sh pattern. CI green, script already executed successfully on all 155 affected portrait videos. Non-blocking nits (gcloud config trap, PROBE_FAIL marked as done) noted in conversation but not worth blocking for a completed one-off.

PR #82, opened 2026-04-16
fix: honor NIP-98 viewer auth for age-gated media

rabble:
## Summary
- accept NIP-98 HTTP auth as viewer auth on media and list routes instead of silently downgrading invalid auth to anonymous
- change `age_restricted` media policy to serve any authenticated viewer while preserving `restricted` as the owner/admin-only shadow-ban path
- keep authenticated and admin media responses off shared caches across blobs, thumbnails, HLS, VTT, audio, and quality variants

## Test Plan
- [x] `cargo test --lib`
- [x] `cargo check --target wasm32-wasi`

### DIFF
--- a/cloud-run-transcoder/src/main.rs
+++ b/cloud-run-transcoder/src/main.rs
@@ -147,7 +147,7 @@ impl Config {
             transcription_retry_total_ms: parse_value(
                 &mut lookup,
                 "TRANSCRIPTION_RETRY_TOTAL_MS",
-                30_000,
+                180_000,
             )
             .max(1_000),
         }
@@ -2332,7 +2332,7 @@ async fn transcribe_via_gemini(
         .post(&url)
         .bearer_auth(&access_token)
         .json(&body)
-        .timeout(std::time::Duration::from_secs(30))
+        .timeout(std::time::Duration::from_secs(120))
         .send()
         .await
         .map_err(|e| {
--- a/docs/superpowers/plans/2026-04-16-media-age-auth-debug.md
+++ b/docs/superpowers/plans/2026-04-16-media-age-auth-debug.md
@@ -1,0 +1,107 @@
+# Media Age Auth Debug Implementation Plan
+
+> **For agentic workers:** REQUIRED: Use superpowers:subagent-driven-development (if subagents available) or superpowers:executing-plans to implement this plan. Steps use checkbox (`- [ ]`) syntax for tracking.
+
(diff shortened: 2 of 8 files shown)


src/main.rs
```diff
@@ -2165,6 +2164,7 @@ fn handle_head_audio(path: &str) -> Result<Response> {
             add_audio_response_headers(
                 &mut resp,
                 &hash,
+                metadata.status.requires_private_cache(),
```
NotThatKindOfDrLiz, line 2167: This HEAD path should go through the same viewer auth/access policy as GET audio. Right now `handle_head_audio` only receives `path`, so it cannot call `viewer_pubkey(&req)?` or evaluate admin auth. That leaves age-restricted audio HEAD inconsistent with GET: valid authenticated viewers can be allowed by GET, but HEAD still evaluates as anonymous.

Can we pass `Request` into this handler, call `viewer_pubkey(&req)?`, compute `is_admin`, and then use `metadata.access_for(requester_pk.as_deref(), is_admin)` here?

docs/superpowers/plans/2026-04-16-media-age-auth-debug.md
```diff
@@ -0,0 +1,107 @@
+# Media Age Auth Debug Implementation Plan
```
NotThatKindOfDrLiz, line 1: Please remove this implementation plan from the PR. It’s useful development scaffolding, but it will go stale quickly once merged to `main`.

cloud-run-transcoder/src/main.rs
```diff
@@ -147,7 +147,7 @@ impl Config {
             transcription_retry_total_ms: parse_value(
                 &mut lookup,
                 "TRANSCRIPTION_RETRY_TOTAL_MS",
-                30_000,
+                180_000,
```
NotThatKindOfDrLiz, line 150: This timeout change looks unrelated to the NIP-98/media auth fix. Since it’s already logically separate, please split it into its own PR unless this auth fix depends on it.

NotThatKindOfDrLiz (changes requested):
The core fix looks right: invalid viewer auth now fails instead of silently becoming anonymous, and age-restricted media is moving to authenticated-viewer access as intended.

Requesting changes for one route consistency issue and one PR hygiene item: wire `HEAD .audio.m4a` through viewer auth, and remove the checked-in implementation plan. I’d also prefer splitting the unrelated transcoder timeout change into its own PR.

PR #85, opened 2026-04-16
feat(admin): creator-delete DELETE action on /admin/api/moderate

mbradley:
## Summary

Blossom side of per-video delete end-to-end enforcement. Wires a `DELETE` action into the `/admin/moderate` webhook handler with a `perform_physical_delete` helper and an `ENABLE_PHYSICAL_DELETE` safety flag.

**Issue:** divinevideo/divine-mobile#3102
**Rollout comment:** https://github.com/divinevideo/divine-mobile/issues/3102#issuecomment-4264760572
**Rollout plan:** [`support-trust-safety/docs/rollout/2026-04-16-creator-delete-rollout.md`](https://github.com/divinevideo/support-trust-safety/blob/working-draft/docs/rollout/2026-04-16-creator-delete-rollout.md)
**Sibling PR:** divinevideo/divine-moderation-service#92 (caller pipeline)

## What's in this PR

- `perform_physical_delete` helper in `src/delete_policy.rs` composing `soft_delete_blob` (status flip, index cleanup, VCL purge) + GCS byte destruction (derived audio, main blob, thumbnail/HLS/VTT artifacts, second VCL purge)
- DELETE special-case branch in `handle_admin_moderate` webhook handler (`src/main.rs`) with `ENABLE_PHYSICAL_DELETE` flag check
- DELETE branch also in `handle_admin_moderate_action` (`src/admin.rs`) as secondary admin-UI path
- Optional `reason` field on `ModerateRequest` for audit log enrichment
- Two main.rs functions promoted to `pub(crate)`: `cleanup_derived_audio_for_source`, `delete_blob_gcs_artifacts`
- `ENABLE_PHYSICAL_DELETE` in `config-store-data.json` (default false) + README docs

## Key design decisions

- **Composes, doesn't extract.** `perform_physical_delete` calls `soft_delete_blob` first (proven helper), then adds byte destruction. Does NOT extract from `execute_vanish` (different semantics: vanish deletes metadata row, creator-delete preserves it for audit).
- **`handle_admin_force_delete` unchanged.** Admin DMCA stays soft-delete by design. Creator-delete physical removal is a separate code path.
- **Flag gates byte destruction only.** Flag off = soft-delete (status Deleted, bytes preserved). Flag on = soft-delete + GCS removal + Fastly Purge. Both paths write audit logs.
- **Honest response.** `physical_deleted` reflects actual success/failure of `perform_physical_delete`. If soft-delete fails inside the helper, the error propagates (blob was never deleted).
- **Audit before destruction.** `write_audit_log` fires before any destructive call, with `action: "creator_delete"` and `actor: metadata.owner`.

### Build verification (done)
- [x] `cargo build --target wasm32-wasip1 --release` succeeds (15 pre-existing warnings, none from our changes)
- [x] `cargo check --tests --locked` succeeds

### Local e2e (done)
- [x] MinIO + Blossom via `docker compose -f docker-compose.local.yml` + `fastly compute serve`
- [x] Flag off: `POST /admin/moderate` with `action: "DELETE"` returns 200 + `physical_delete_skipped: true`. Blob serves 404 (status Deleted).
- [x] Flag on: same call returns 200 + `physical_deleted: true`. Physical delete cascade executed.
- [x] Webhook auth: `Bearer dev-secret-for-testing` accepted, missing/invalid rejected
- [x] BAN action still works unchanged: `status: "banned", success: true`
- [x] Admin DMCA endpoint (`/admin/api/delete`) still works unchanged: `preserved: true, deleted: true`
- [x] Audit log attempted (Cloud Run backend not available locally, expected fire-and-forget)
- [x] VCL purge skipped locally (no `fastly_api_token`, expected log message)

### Production validation (per rollout plan)
- [ ] Step 1: deploy with flag off, verify DELETE accepted (not "Unknown action")
- [ ] Step 6: flip flag, verify physical byte removal on next creator-delete

### Code review (done)
- [x] Three rounds of self-review including cross-repo integration contract
- [x] All findings addressed: soft-delete failure propagates as error, GCS delete failure logged, response accuracy for physical_deleted, config access pattern documented

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ Client → Fastly Compute (Rust WASM) → GCS (blobs) + Fastly KV (metadata)
 rustup target add wasm32-wasi
 ```
 
-### Configure secrets
+### Configure secrets and config
 
 1. Create a GCS bucket with HMAC credentials
 2. Set up Fastly stores:
@@ -65,6 +65,12 @@ fastly config-store create --name blossom_config
 # Create secret store with GCS HMAC credentials
 fastly secret-store create --name blossom_secrets
 ```
+
+**Config store flags:**
+
+| Key | Description | Default |
+|-----|-------------|---------|
+| `ENABLE_PHYSICAL_DELETE` | When `"true"`, creator-delete actions via `/admin/api/moderate` physically remove bytes from GCS and purge edge caches. When `"false"`, status flip only (bytes preserved). Admin DMCA via `/admin/api/delete` is unconditionally soft-delete regardless of this flag. | `"false"` |
 
 ### Local development
 
--- a/config-store-data.json
+++ b/config-store-data.json
@@ -1,5 +1,6 @@
 {
   "gcs_bucket": "divine-blossom-local",
   "gcs_project_id": "local",
-  "local_mode": "true"
+  "local_mode": "true",
+  "ENABLE_PHYSICAL_DELETE": "false"
 }
--- a/docs/superpowers/plans/2026-04-16-creator-delete-action-plan.md
+++ b/docs/superpowers/plans/2026-04-16-creator-delete-action-plan.md
@@ -1,175 +1,112 @@
-# Blossom `DELETE` Action Implementation Plan
+# Blossom `DELETE` Action Implementation Plan (revised)
 
 > **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
 
-**Goal:** Wire a `DELETE` action into Blossom's `/admin/api/moderate` endpoint that dispatches to a shared physical-delete helper extracted from the existing admin DMCA cascade. Gate the destructive step on a new `ENABLE_PHYSICAL_DELETE` config flag (default off) so the first production deploy is inert.
+**Goal:** Wire a `DELETE` action into Blossom's `/admin/api/moderate` endpoint that composes `soft_delete_blob` (stops serving) with physical GCS byte removal (when the `ENABLE_PHYSICAL_DELETE` flag is on). New helper lives in `delete_policy.rs` alongside the existing soft-delete helper.
 
-**Architecture:** Extract `perform_physical_delete(hash, actor, reason, legal_hold)` from `handle_admin_force_delete` into a shared helper. Both admin DMCA (unconditional) and creator-delete (flag-gated) call it. Creator-delete ingress stays on the existing `/admin/api/moderate` endpoint so moderation-service's single call pattern is preserved.
+**Architecture:** `perform_physical_delete` = `soft_delete_blob` + byte destruction. DELETE handled as a special-case branch in `handle_admin_moderate_action` (not a new arm in the bare-status-flip match). Admin DMCA endpoint and GDPR vanish are untouched.
 
-**Tech Stack:** Rust on Fastly Compute (WASM), `fastly` crate, `serde_json`. Tests via inline `#[cfg(test)] mod tests` — standard Rust pattern. Local dev via MinIO + `fastly compute serve`.
+**Tech Stack:** Rust on Fastly Compute (WASM target `wasm32-wasip1`), `fastly` crate, `serde_json`. Tests via `cargo check --tests --locked` (the edge crate's FFI symbols don't link for `cargo test` on host). Local e2e via MinIO + `fastly compute serve`.
 
-**Spec:** `docs/superpowers/specs/2026-04-16-creator-delete-action-design.md` on this branch.
+**Spec:** `docs/superpowers/specs/2026-04-16-creator-delete-action-design.md` (revised).
 
 ---
 
 ## File Structure
 
 **Files to modify:**
-- `src/main.rs` — extract `perform_physical_delete` helper (~55 lines moved from `handle_admin_force_delete`), update `handle_admin_force_delete` to call it.
-- `src/admin.rs` — add `reason` field to `ModerateRequest` struct, wire `DELETE` action dispatch with `ENABLE_PHYSICAL_DELETE` flag check.
-- `config-store-data.json` — add `ENABLE_PHYSICAL_DELETE = "false"` for local dev.
+- `src/delete_policy.rs` — add `perform_physical_delete` (new public function, ~20 lines)
+- `src/admin.rs` — add `reason` field to `ModerateRequest`; add DELETE special-case branch in `handle_admin_moderate_action` (~40 lines)
+- `config-store-data.json` — add `ENABLE_PHYSICAL_DELETE = "false"` for local dev
+- `README.md` — document the flag
 
-**Files to create:**
-- None. Tests live alongside production code in `#[cfg(test)]` modules within the modified files.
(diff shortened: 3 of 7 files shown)


NotThatKindOfDrLiz (changes requested):
Thanks for the implementation and the later fixes on soft-delete failure handling. I think this is close, but I would not merge it until two correctness issues are addressed.

## Requested changes before merge

### 1. Remove the duplicated creator-delete implementation

The `DELETE` branch now exists in two places:

- `src/main.rs` in `handle_admin_moderate` for `/admin/moderate`
- `src/admin.rs` in `handle_admin_moderate_action` for `/admin/api/moderate`

Those branches perform the same policy work but already differ in response shape, CORS handling, request parsing, and config access. Since this is a cross-service contract with moderation-service, that duplication is risky.

Please extract the creator-delete policy into one shared helper and have both handlers call it. A reasonable shape would be something like:

```rust
pub struct CreatorDeleteOutcome {
    pub physical_delete_enabled: bool,
    pub physical_deleted: bool,
}

pub fn handle_creator_delete(
    hash: &str,
    metadata: &BlobMetadata,
    reason: &str,
    physical_delete_enabled: bool,
) -> Result<CreatorDeleteOutcome>
```

The handlers should be thin adapters: parse/authenticate, load metadata, call the shared helper, and serialize a consistent response. Ideally both return the same core fields: `success`, `sha256`, `old_status`, `new_status`, `physical_deleted`, and `physical_delete_skipped`.

### 2. Make `physical_deleted` reflect actual main-blob deletion success

The later commits improved the response when `soft_delete_blob` fails, which is good. However, `perform_physical_delete` still swallows the main GCS delete failure:

```rust
if let Err(e) = crate::storage::delete_blob(hash) {
    eprintln!("[PHYSICAL-DELETE] storage::delete_blob failed ... Bytes may remain on GCS.");
}
```

The helper then returns `Ok(())`, so callers can still emit `physical_deleted: true` even though the primary GCS object remains. For a compliance-sensitive creator-delete path, that is not an acceptable success signal.

Please either:

- propagate the main blob deletion error so `physical_deleted` remains false and the caller sees a failure, or
- return a structured outcome that distinguishes `soft_deleted`, `main_blob_deleted`, artifact cleanup status, and overall `physical_deleted`.

My recommendation: treat main blob deletion failure as an `Err`; keep artifact cleanup best-effort if that matches existing behavior.

### 3. Do not audit success before the delete outcome is known

Both DELETE branches write `creator_delete` audit before the soft-delete/physical-delete work has succeeded. If `soft_delete_blob` fails afterward, the audit trail can imply a delete happened when it did not.

Please move the success audit after the status flip succeeds, or add explicit failure audit entries with an outcome field. For this PR, moving the success audit after successful soft delete is probably enough.

### 4. Share config access instead of inlining `ConfigStore::open`

`main.rs` currently opens `ConfigStore::open("blossom_config")` directly because `admin::get_config` is private. Please make the config accessor shared (`pub(crate)` or moved to a small common helper) and use it from both paths. That avoids duplicating the store name and keeps future config behavior centralized.

## Non-blocking follow-ups

I am going to file separate issues for test coverage, request correlation IDs, the rollout sweep, response-contract documentation, and repo policy for `docs/superpowers` scaffolding. Those should not block this PR once the correctness items above are addressed.

Please rerun at least:

```bash
cargo check --tests --locked --target wasm32-wasip1
```

A short manual validation note for flag-off and flag-on creator-delete behavior would also be useful before merge.

mbradley:
Three commits addressing the review. Summary:

**c6bdb0f — first pass on the four correctness items.** Extracted `handle_creator_delete` into `delete_policy.rs`. Both handlers call it as thin adapters. `perform_physical_delete` propagated `Err`. Audit moved to after success. `admin::get_config` is `pub(crate)`.

**747d147 — tried structured outcome.** Self-review found the first pass created an audit gap when soft succeeded but byte-delete failed (Err returned early, no audit written). Restructured to return `Ok(outcome)` with `soft_deleted`/`physical_deleted` fields so partial state was representable, and encoded outcome in the audit action (`creator_delete_partial` / `creator_delete_failed`).

**a380c2e — reverted to Err-propagation (Liz's stated recommendation)** after pressure-testing the tradeoff in this system's context. Kept the intent of the second commit (no audit gap) but used a different mechanism.

## Why we ended up where Liz recommended

Structured-outcome is a defensible choice in isolation, but in this system `notifyBlossom` in moderation-service reads only `response.ok` today. Shipping partial-success as HTTP 200 would silently mark byte-delete failures as accepted in the `creator_deletions` D1 row, relying on the #90 validation-window sweep to eventually reconcile. Err-propagation returns 5xx on byte-delete failure, moderation-service's existing retry logic catches it, and the pending D1 row stays visible in the dashboard today. Loud signal beats silent partial success for a compliance-sensitive path when the primary caller does not consume the partial signal. Doing the structured-outcome path "properly" would require moderation-service work (new state, byte-only retry, dashboard surfacing) that is not justified given byte-delete failures are rare and retry-whole-op converges at negligible extra cost.

## Closing the audit gap without schema changes

Two writes per attempt, bracketing the helper call:

- `creator_delete_attempt` fires before the helper.
- `creator_delete` fires after the helper returns `Ok`.

A failed attempt leaves an attempt entry with no paired success. Operators can query `action = 'creator_delete_attempt'` and diff against `creator_delete` entries by `sha256` to enumerate failures. No audit schema change needed.

## Final response shape (both handlers)

`{success, sha256, old_status, new_status: "deleted", physical_deleted, physical_delete_skipped}` — matches the fields Liz listed.

## Manual validation

**Flag-off (ENABLE_PHYSICAL_DELETE unset or not "true"):**
- POST with valid bearer and a known `sha256`.
- Expect 200, `physical_deleted: false`, `physical_delete_skipped: true`, `new_status: "deleted"`.
- GET the blob → 404.
- GCS main bytes remain.
- Two audit entries: `creator_delete_attempt` then `creator_delete`.

**Flag-on, full success (ENABLE_PHYSICAL_DELETE="true"):**
- Same request.
- Expect 200, `physical_deleted: true`, `physical_delete_skipped: false`.
- Blob stops serving, GCS main bytes gone.
- Two audit entries: `creator_delete_attempt` then `creator_delete`.

**Flag-on, byte-delete fails (simulate via GCS outage or bad key):**
- Expect 5xx.
- Blob status is `Deleted` on Blossom (soft-delete already applied).
- GCS bytes remain.
- One audit entry: `creator_delete_attempt` (no paired success).
- Caller retry converges: `storage::delete_blob` treats missing object as success, `soft_delete_blob` is no-op on already-`Deleted` state.

**Failure paths:**
- Bad sha256 → 400 (before audit; unchanged).
- Missing blob → 404 (before audit; unchanged).
- Soft-delete fails → 5xx, one `creator_delete_attempt` audit entry.

Ready for re-review.

NotThatKindOfDrLiz (approved):
Approved. The prior blockers are addressed.

The creator-delete policy is now centralized in `handle_creator_delete`, both moderation handlers return the same response shape, main GCS delete failures propagate instead of allowing `physical_deleted: true`, and the attempt/success audit bracketing gives us a useful failure signal without changing the audit schema.

CI is green. The remaining retry/idempotency coverage and operational sweep work are already tracked as follow-ups, so I’m comfortable merging this once the branch is otherwise ready.

PR #92, opened 2026-04-17
feat(obs): thread request correlation IDs through moderate and creator-delete logs

### LINKED ISSUE #86: Add request correlation IDs to moderation and delete logs
Follow-up from PR #85 review. This should not block PR #85 once the requested correctness fixes land.

## Scope

Add or propagate a request/correlation ID through moderation and delete flows so retries and partial failures can be traced across logs.

Relevant paths:

- `/admin/moderate`
- `/admin/api/moderate`
- creator-delete policy helper
- physical delete helper
- Fastly purge logs
- audit log entries where appropriate

## Acceptance criteria

- Prefer an incoming request ID header if present.
- Generate a short stable-per-request ID when one is not present.
- Include the ID in relevant `eprintln!` lines for delete, purge, and failure paths.
- Document the header name expected from upstream callers such as moderation-service.

mbradley:
Closes #86.

Stacked on #85 (`feat/creator-delete-action`) — do not merge until #85 lands so the diff stays clean.

## Summary

Adds request correlation IDs to the moderate and creator-delete log paths so retries and partial failures across `/admin/moderate` and `/admin/api/moderate` can be traced in a single grep.

## What changed

- New `src/req_id.rs` helper. Priority: `X-Request-Id` header > leading segment of `cf-ray` > generated 12-hex-char ID from nanosecond clock.
- `handle_creator_delete` gains a `req_id: &str` argument; its internal byte-delete failure log now includes the ID.
- `handle_admin_moderate` (main.rs) and `handle_admin_moderate_action` (admin.rs) extract the ID at entry and prefix their `[ADMIN]` / `[CREATOR-DELETE]` log lines with `[req=<id>]`.
- `X-Request-Id` added to the CORS allowed-headers list in all three CORS sites.
- README section documenting the header contract and fallback behavior.

## Scope decisions

- **`purge_vcl_cache` signature unchanged.** 11 call sites, several in background / non-request paths. Its `[PURGE]` logs already include the sha256 as the surrogate key, which correlates back to request-context logs via that shared value when purging is triggered from a moderate/delete path. Threading `req_id` everywhere wasn't worth the ripple for this pass. Follow-up material if operators want tighter purge tracing.
- **Audit entries unchanged.** Adding a req_id column would require schema work; the spec's "audit log entries where appropriate" was optional. The paired `creator_delete_attempt` / `creator_delete` entries from #85 already let operators reconstruct a trace via sha256 + timestamp.

## Acceptance criteria

- [x] Prefer incoming `X-Request-Id` header.
- [x] Generate a short stable-per-request ID when absent.
- [x] Include the ID in `eprintln!` lines for delete and failure paths (`[CREATOR-DELETE]`, `[ADMIN]`).
- [x] Document the header name expected from upstream callers (README).

### DIFF
--- a/README.md
+++ b/README.md
@@ -255,6 +255,26 @@ to the owner or an admin. Blossom does not currently read any hosted-session
 age-verification claim or external viewer adult-verification service when
 serving media.
 
+## Request correlation
+
+Admin and moderation endpoints accept an `X-Request-Id` header and include
+its value in related log lines so retries and partial failures can be traced
+across stderr.
+
+- If the caller sends `X-Request-Id`, its first 16 characters are used.
+- If absent, the leading segment of the Cloudflare-provided `cf-ray`
+  header is used (free correlation with Cloudflare edge logs).
+- If neither is present, a short hex ID is generated from the nanosecond
+  clock.
+
+Upstream services integrating with Blossom (e.g. `divine-moderation-service`
+for creator-delete) are encouraged to forward a stable `X-Request-Id`
+across their retry loops so both sides share the same correlation token.
+Log lines are prefixed with `[req=<id>]`; the `[PURGE]` logs emitted by
+the Fastly cache-purge path include the blob `sha256` as their surrogate
+key, which serves as the cross-reference when purging is triggered from
+a moderate/delete request.
+
 ## License
 
 MIT
--- a/src/admin.rs
+++ b/src/admin.rs
@@ -806,7 +806,7 @@ fn set_admin_blob_response_headers(
     resp.set_header("Access-Control-Allow-Methods", "GET, HEAD, OPTIONS");
     resp.set_header(
         "Access-Control-Allow-Headers",
-        "Authorization, Content-Type, Range",
+        "Authorization, Content-Type, Range, X-Request-Id",
     );
 }
 
@@ -851,6 +851,8 @@ struct RestoreRequest {
 
 pub fn handle_admin_moderate_action(mut req: Request) -> Result<Response> {
     validate_admin_auth(&req)?;
+
+    let req_id = crate::req_id::for_request(&req);
 
     let body = req.take_body().into_string();
     let moderate_req: ModerateRequest = serde_json::from_str(&body)
@@ -902,11 +904,12 @@ pub fn handle_admin_moderate_action(mut req: Request) -> Result<Response> {
             &metadata,
             reason,
             physical_delete_enabled,
+            &req_id,
         )
         .map_err(|e| {
             eprintln!(
-                "[CREATOR-DELETE] handle_creator_delete failed for {}: {}",
-                moderate_req.sha256, e
+                "[req={}] [CREATOR-DELETE] handle_creator_delete failed for {}: {}",
+                req_id, moderate_req.sha256, e
             );
             e
         })?;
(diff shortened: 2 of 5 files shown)


rabble (approved):
## Overview

Small, well-scoped observability PR: adds `src/req_id.rs` (header → cf-ray → nanosecond-hex fallback), threads a `req_id` through `/admin/moderate` (main.rs) and `/admin/api/moderate` (admin.rs) plus `handle_creator_delete`, and adds `X-Request-Id` to three CORS allow-lists. README contract is clear.

**Diff scope note:** `baseRefName=main` so this diff includes #85's creator-delete code, and GitHub reports `mergeable: CONFLICTING`. Rebase after #85 lands so the diff stays focused on correlation-ID threading.

## Issues / Suggestions

### 1. Log injection via user-controlled `X-Request-Id` (please address before merge)

`src/req_id.rs` trims + truncates but does not filter characters. A post-auth caller to `/admin/*` can send `X-Request-Id: \x1b[31m\n[ADMIN] fake line` and inject ANSI/newlines into stderr. Suggest restricting to `[A-Za-z0-9_-]` and applying the same filter to the `cf-ray` segment:

```rust
fn sanitize(s: &str) -> String {
    s.chars()
        .filter(|c| c.is_ascii_alphanumeric() || *c == '-' || *c == '_')
        .take(MAX_LEN)
        .collect()
}
```

### 2. Generated-ID collision risk on concurrent requests

`generate()` masks `ns` to 48 bits. On Fastly Compute WASM, `SystemTime::now()` resolution is often coarser than true nanoseconds — two requests on the same POP in the same tick will collide. Only fires when both header fallbacks are absent (rare), but a tiny `AtomicU64` counter XORed into the low bits would eliminate the risk at zero cost.

### 3. DELETE branch duplicated between `handle_admin_moderate` and `handle_admin_moderate_action`

~40 near-identical lines in `src/main.rs` and `src/admin.rs`, and the two adapters are already drifting subtly (main.rs calls `add_cors_headers`, admin.rs doesn't; `reason` is pulled from different places). A shared `respond_creator_delete(sha256, metadata, reason, enabled, req_id) -> Result<Response>` helper would prevent drift. Not blocking.

### 4. Test coverage gaps in `req_id.rs`

Current tests cover only `truncate` and `generate`. Consider adding:
- `x-request-id` wins over `cf-ray` when both present
- `cf-ray` segment extraction (splits on `-`)
- Whitespace-only / empty header falls through to next source

Cheap to add with a small seam, materially raises confidence in the priority contract.

### 5. Minor nits

- README says "first 16 characters" but generated IDs are 12 hex chars — worth a one-liner to prevent operator confusion when grepping.
- `pub(crate)` escalations in `main.rs` (`cleanup_derived_audio_for_source`, `delete_blob_gcs_artifacts`, `get_config`) are fine for now; longer-term these utilities probably want to leave `main.rs`.

## Security / Performance

- Log injection (item 1) is the only security concern; auth gate limits blast radius but doesn't eliminate it.
- No secrets logged; `req_id` alone is not PII; 16-char cap keeps log lines bounded.
- Zero new syscalls / KV reads / backend calls; one `String` + `format!` on the fallback path.

## Recommendation

**Approving.** Correlation-ID helper is small, useful, and well-documented. Please address item 1 (sanitize the header) before merge and rebase once #85 lands. Items 2–5 are follow-up material.

rabble:
Pushed 65c64e6 to address review item 1 (log injection). Swapped trim→`sanitize` that filters to `[A-Za-z0-9_-]` before the 16-char cap, applied to both `X-Request-Id` and the `cf-ray` segment. Added tests for the injection cases (newlines, ANSI escapes, whitespace). `cargo check --tests --locked --target wasm32-wasip1` clean.

rabble:
Rebased onto main (#85 landed). Force-pushed; branch is now 3 commits / +159-20 covering only the req-id threading. Status: `MERGEABLE`.

Issue #95, opened 2026-04-17
Status strings in admin responses use Debug, not serde (age_restricted → agerestricted)


rabble:
## Problem

Admin response handlers render `BlobStatus` via `format!("{:?}", status).to_lowercase()`, which bypasses the `#[serde(rename = "age_restricted")]` attribute on `BlobStatus::AgeRestricted` (defined in `src/blossom.rs`). As a result, the `old_status` / `new_status` fields emit `"agerestricted"` (no underscore) instead of the canonical `"age_restricted"` used elsewhere in the API.

## Affected call sites

- `src/main.rs` — creator-delete response in `handle_admin_moderate` (line ~4815)
- `src/admin.rs` — creator-delete response in `handle_admin_moderate_action`
- `src/admin.rs` — status-flip response in `handle_admin_moderate_action`
- `src/admin.rs` — restore response in `handle_admin_restore_action`

All four use the same `format!("{:?}", status).to_lowercase()` pattern.

## Impact

- Callers matching `old_status` / `new_status` against canonical status strings (e.g. `"age_restricted"` from GET endpoints that use serde) get a mismatch on the age-gated variant on these four response paths.
- Documented contract in `docs/api/creator-delete-contract.md` had to note the discrepancy: see the "Note: status string rendering" section.
- Other status variants (`active`, `restricted`, `pending`, `banned`, `deleted`) happen to render identically via Debug+lowercase and serde, so this bug is limited to `AgeRestricted`.

## Suggested fix

Replace the `format!("{:?}", ...).to_lowercase()` pattern with a serde-aware rendering helper. One approach:

```rust
fn status_as_api_str(s: BlobStatus) -> &'static str {
    match s {
        BlobStatus::Active => "active",
        BlobStatus::Restricted => "restricted",
        BlobStatus::AgeRestricted => "age_restricted",
        BlobStatus::Pending => "pending",
        BlobStatus::Banned => "banned",
        BlobStatus::Deleted => "deleted",
    }
}
```

Or use `serde_json::to_value(status).as_str()` inside a thin helper. The explicit match is cheaper at runtime and keeps the API-string contract visible.

## Follow-up

Once fixed, remove the "Note: status string rendering" section from `docs/api/creator-delete-contract.md` and update the current-values list back to include `age_restricted`.

Discovered during post-merge review of #94.

sergey3bv:
If you don't mind I will work on the issue

sergey3bv:
[...]

Issue #87, opened 2026-04-17
test(moderation): add creator-delete behavior coverage


NotThatKindOfDrLiz:
Follow-up from PR #85 review. This should not block PR #85 once the requested correctness fixes land.

## Scope

Add automated coverage for creator-delete behavior across the moderation entry points.

Suggested cases:

- `DELETE` with `ENABLE_PHYSICAL_DELETE=false` returns success with `physical_delete_skipped: true` and flips the blob to `Deleted`.
- `DELETE` with `ENABLE_PHYSICAL_DELETE=true` reports `physical_deleted: true` only when the main blob delete succeeds.
- Invalid sha256 returns 400.
- Missing blob returns 404.
- Unknown action behavior remains unchanged.
- `/admin/moderate` and `/admin/api/moderate` keep a consistent response contract.

## Notes

Fastly/Viceroy setup may make full handler tests expensive. If so, prioritize unit coverage around the extracted creator-delete policy helper plus a lightweight response-shape fixture.

PR #97, opened 2026-04-17
test(delete): unit coverage for creator-delete response contract

mbradley:
Refs #87.

## Summary

Adds unit coverage for the creator-delete response contract and extracts a shared `build_creator_delete_response(sha256, outcome)` helper so `/admin/moderate` and `/admin/api/moderate` produce the same response via one code path.

## Coverage

Per Liz's note on #87 ("prioritize unit coverage around the extracted creator-delete policy helper plus a lightweight response-shape fixture"), this PR stays at the unit level and does not spin up Viceroy for full handler tests.

- Response-shape test for flag-off active blob (base case).
- Response-shape test for flag-on active blob with successful byte-delete.
- Response-shape test for idempotent retry on an already-`Deleted` blob.
- Parameterized test pinning `old_status` rendering for every `BlobStatus` variant (`active`, `restricted`, `pending`, `banned`, `deleted`, `agerestricted`).

Handler-level pre-validation paths (`Invalid sha256 → 400`, `missing blob → 404`, `unknown action unchanged`) return before the helper is reached. They're exercised by the rollout plan's Step 1 smoke test and ship without new unit coverage per Liz's Viceroy-cost note.

## Refactor

Both handlers now call `build_creator_delete_response` instead of inlining `serde_json::json!({...})`. Response shape guaranteed identical via a single source. `main.rs` preserves its pre-existing `add_cors_headers` call; `admin.rs` remains CORS-free as before.

## Out of scope

- #95 (serde-aware `BlobStatus` rendering) — the variant test pins current behavior (`"agerestricted"`, no underscore) with an inline comment. When #95 lands, that one assertion flips to `\"age_restricted\"`.

## Review pass

Ran the code-reviewer agent on the diff before opening. Two findings:
- **CORS asymmetry between handlers** (flagged as concern) — verified against origin/main that `main.rs`'s `add_cors_headers` call was pre-existing; the refactor preserved it. No behavior change.
- **Tautological assertions in a truth-table test** — dropped the test per feedback. The readable-enumeration intent it served is already covered by the distinct `response_builder_*` tests.

## Verification

### DIFF
--- a/src/admin.rs
+++ b/src/admin.rs
@@ -3,7 +3,8 @@
 
 use crate::blossom::{BlobMetadata, BlobStatus, GlobalStats, RecentIndex};
 use crate::delete_policy::{
-    handle_creator_delete, parse_restore_status, restore_soft_deleted_blob,
+    build_creator_delete_response, handle_creator_delete, parse_restore_status,
+    restore_soft_deleted_blob,
 };
 use crate::error::{BlossomError, Result};
 use crate::metadata::{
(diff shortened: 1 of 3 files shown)


NotThatKindOfDrLiz (approved):
Approved.

The shared `build_creator_delete_response` helper removes the remaining response-shape drift risk between `/admin/moderate` and `/admin/api/moderate`, and the new unit tests pin the current creator-delete response contract in a focused, low-cost way.

Changing the PR body to `Refs #87` was the right correction. The remaining helper/pre-validation coverage is now tracked separately in #98, so I’m comfortable landing this as the response-contract portion of the work.

Issue #98, opened 2026-04-22
Add remaining creator-delete helper and pre-validation coverage


NotThatKindOfDrLiz:
Follow-up from PR #97. PR #97 covered the shared creator-delete response builder and response-contract shape, but it did not add automated coverage around `handle_creator_delete` itself or the handler pre-validation paths.

## Scope

Add the remaining automated coverage for creator-delete behavior where it is practical without over-investing in Viceroy-heavy tests.

Priority targets:

- Unit coverage around `handle_creator_delete` for the two successful helper outcomes:
  - `physical_delete_enabled=false` returns `Ok` with `physical_deleted=false`
  - `physical_delete_enabled=true` and successful byte delete returns `Ok` with `physical_deleted=true`
- If a low-cost seam exists, add targeted coverage for the byte-delete failure path to pin the current contract: helper returns `Err` after soft-delete has already applied.
- Add lightweight coverage for pre-validation paths if they can be tested cheaply:
  - invalid sha256 -> 400
  - missing blob -> 404
  - unknown action unchanged

## Notes

- Full end-to-end handler tests via Viceroy remain optional if the setup cost is high.
- It is acceptable to split helper coverage and pre-validation coverage if they need different test strategies.
- Keep the response-shape tests added in PR #97 as the source for cross-handler JSON contract parity.

PR #99, opened 2026-04-23
test: cover handle_creator_delete outcomes via trait-based DI

mbradley:
Refs #98.

## Summary

Compile-checked unit coverage on `handle_creator_delete` itself, per the priority targets in #98. Needed a light refactor because the helper's five side-effectful calls (`soft_delete_blob`, `crate::cleanup_derived_audio_for_source`, `crate::storage::delete_blob`, `crate::delete_blob_gcs_artifacts`, `crate::purge_vcl_cache`) go straight into Fastly KV / GCS / VCL, which makes native unit testing impossible without Viceroy.

## Refactor

- New `CreatorDeleteOps` trait abstracting the five side effects
- New `DefaultCreatorDeleteOps` impl that forwards to the existing free functions (production behavior unchanged)
- `handle_creator_delete`'s public signature is unchanged. Internally it delegates to `handle_creator_delete_with_ops(..., &DefaultCreatorDeleteOps)`
- The DI seam is internal to the crate (`pub(crate)`), matching its test-only role
- No changes to `admin.rs` or `main.rs`, so the wire contract pinned in PR #97 is untouched

## Tests added (alongside the existing response-builder tests)

- **flag-off path:** only `soft_delete` is called; `physical_deleted` is false
- **flag-on success path:** full sequence runs in order (`soft_delete`, `cleanup_derived_audio`, `delete_blob`, `delete_blob_gcs_artifacts`, `purge_vcl_cache`); `physical_deleted` is true
- **flag-on byte-delete failure path:** `soft_delete` and `cleanup_derived_audio` ran, `delete_blob` returned `Err`, no GCS artifact cleanup or VCL purge happened, and the error propagates. This pins the partial-state invariant the helper docstring describes (soft-delete applied before byte-delete failed, caller gets loud `Err`).
- **soft-delete failure path:** short-circuits before any physical-delete ops; error propagates.
- **`old_status` capture:** the outcome's `old_status` is the value from `metadata.status` at entry, covered across every `BlobStatus` variant.

Mock impl uses `RefCell<Vec<&'static str>>` to record call order and `Option<BlossomError>` to preset fallible-op outcomes.

## Scope I intentionally did not take

Pre-validation coverage (invalid sha256, missing blob, unknown action). #98 remains open for that remaining coverage and for any future runnable/Viceroy-based execution path. Invalid sha256 is a pure string check buried in the handler; missing blob needs a KV seam that costs more than a Viceroy setup.

## Test plan

- [x] `cargo check --all-targets` clean
- [x] `cargo clippy --all-targets` clean (no new warnings on `delete_policy.rs`)
- [x] Existing `delete_policy` tests still present and compile
- Note: tests run natively in CI only under the `cloud-run-upload/` Cargo manifest. Tests in `src/delete_policy.rs` (both the ones added in #97 and these new ones) require linking Fastly FFI symbols and are compile-checked in CI today; they run fully under Viceroy or on the Fastly platform.

### DIFF
--- a/src/delete_policy.rs
+++ b/src/delete_policy.rs
@@ -79,6 +79,40 @@ pub struct CreatorDeleteOutcome {
     pub physical_deleted: bool,
 }
 
+/// Side-effects used by `handle_creator_delete`. The default impl
+/// (`DefaultCreatorDeleteOps`) forwards to the crate-level functions that
+/// talk to Fastly KV, GCS, and the VCL cache. Tests substitute a mock so
+/// `handle_creator_delete` can be exercised natively without Viceroy.
+pub trait CreatorDeleteOps {
+    fn soft_delete(&self, hash: &str, metadata: &BlobMetadata, reason: &str) -> Result<()>;
+    fn cleanup_derived_audio(&self, hash: &str);
+    fn delete_blob(&self, hash: &str) -> Result<()>;
+    fn delete_blob_gcs_artifacts(&self, hash: &str);
+    fn purge_vcl_cache(&self, hash: &str);
+}
+
+/// Production-side implementation. Forwards to the real Fastly-backed
+/// functions. Only used inside `handle_creator_delete`'s default path.
+pub struct DefaultCreatorDeleteOps;
+
+impl CreatorDeleteOps for DefaultCreatorDeleteOps {
+    fn soft_delete(&self, hash: &str, metadata: &BlobMetadata, reason: &str) -> Result<()> {
+        soft_delete_blob(hash, metadata, reason, false)
+    }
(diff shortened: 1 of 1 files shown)


NotThatKindOfDrLiz (approved):
Approved.

The core DI refactor is sound, and the two review concerns are now addressed: the `CreatorDeleteOps` seam is crate-private, and the PR/body scope now correctly treats this as compile-checked helper coverage while leaving #98 open for any remaining runnable or pre-validation coverage.

The public `handle_creator_delete` contract stays unchanged, `admin.rs`/`main.rs` remain untouched, and CI is green. Good to merge.

PR #100, opened 2026-04-24
Fixing serialize admin BlobStatus using canonical API names

### LINKED ISSUE #95: Status strings in admin responses use Debug, not serde (age_restricted → agerestricted)

sergey3bv:
Should close #95

### DIFF
--- a/docs/api/creator-delete-contract.md
+++ b/docs/api/creator-delete-contract.md
@@ -44,18 +44,12 @@ Field semantics:
 |---|---|---|
 | `success` | bool | Always `true` when HTTP 200. |
 | `sha256` | string | Echoes the request blob hash. |
-| `old_status` | string | Lowercase `BlobStatus` the blob was in before this call. Current values: `active`, `restricted`, `agerestricted`, `pending`, `banned`, `deleted`. See [note on status rendering](#note-status-string-rendering). |
+| `old_status` | string | Lowercase `BlobStatus` the blob was in before this call. Current values: `active`, `restricted`, `age_restricted`, `pending`, `banned`, `deleted`. |
 | `new_status` | string | Always `"deleted"` on success. |
 | `physical_deleted` | bool | `true` only when the main GCS blob delete succeeded. `false` when the flag is off *or* when physical delete wasn't attempted. |
 | `physical_delete_skipped` | bool | `true` when `ENABLE_PHYSICAL_DELETE` was off; `false` when it was on. |
 
 Callers can rely on these fields without reading Blossom source. Additional fields may be added; clients must ignore unknown fields.
-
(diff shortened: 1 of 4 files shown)


NotThatKindOfDrLiz (approved):
Approved.

This is the right fix for #95. Moving status rendering to `BlobStatus::as_api_str()` removes the remaining mismatch between the admin responses and the canonical API names, and updating the shared creator-delete response builder means both moderation entry points pick up the fix through one path.

The diff stays tight, the contract doc is updated to match, and CI is green. Good to merge.

PR #103, opened 2026-04-28
test(moderation): add shared validation and action-mapping unit tests

mbradley:
## Summary

- Extract `validate_sha256_format()` from the two moderation handlers into a shared pure helper in `delete_policy.rs`
- Extract endpoint-specific action mappers into shared pure helpers in `delete_policy.rs`:
  - `map_admin_api_action()` preserves the existing `/admin/api/moderate` alias set and error behavior
  - `map_webhook_moderate_action()` preserves the existing `/admin/moderate` alias set and error behavior
- Add unit tests covering SHA-256 validation, accepted aliases, rejected aliases, and unknown-action behavior for both ingress paths
- Document remaining coverage gaps: missing-blob 404 still needs an integration/Viceroy path, and runnable native execution of these tests is tracked separately

**Note:** These tests are compile-verified via `cargo check --tests --target wasm32-wasip1` but do not execute natively yet because of the Fastly FFI link constraint. Issue #104 tracks the `blossom-core` extraction needed to run them under normal `cargo test`.

**Stack:** #105 depends on this PR. Retarget #105 to `main` before merging this.

## Test plan

- [x] `cargo check --target wasm32-wasip1` passes (no errors, pre-existing warnings only)
- [x] `cargo check --tests --target wasm32-wasip1` passes
- [x] Unit tests pin the existing per-endpoint alias sets and unknown-action behavior for both moderation handlers
- [x] Rebased on current `main` after #99 landed so the crate-private test seam remains intact

Refs #87

### DIFF
--- a/src/delete_policy.rs
+++ b/src/delete_policy.rs
@@ -12,6 +12,37 @@ pub enum DeletePlan {
     UnlinkOnly,
 }
 
+/// Validates that a SHA-256 string is exactly 64 hex characters.
+/// Used by both `/admin/moderate` and `/admin/api/moderate` before any
+/// metadata lookup.
+pub fn validate_sha256_format(sha256: &str) -> Result<()> {
+    if sha256.len() != 64 || !sha256.chars().all(|c| c.is_ascii_hexdigit()) {
+        return Err(BlossomError::BadRequest("Invalid sha256 format".into()));
+    }
+    Ok(())
+}
+
+/// Maps an action string from a moderation request to a `BlobStatus`.
+/// Returns `Err(BadRequest)` for unknown actions. The DELETE action is
+/// handled separately by callers before reaching this function.
+///
+/// Both `/admin/moderate` and `/admin/api/moderate` accept slightly
+/// different action aliases. This function covers the union so both
+/// endpoints can share validation.
+pub fn map_moderate_action(action: &str) -> Result<BlobStatus> {
+    match action.to_uppercase().as_str() {
+        "BAN" | "BLOCK" | "PERMANENT_BAN" => Ok(BlobStatus::Banned),
+        "RESTRICT" | "QUARANTINE" => Ok(BlobStatus::Restricted),
+        "AGE_RESTRICT" | "AGE_RESTRICTED" => Ok(BlobStatus::AgeRestricted),
+        "APPROVE" | "ACTIVE" | "SAFE" => Ok(BlobStatus::Active),
+        "PENDING" => Ok(BlobStatus::Pending),
+        _ => Err(BlossomError::BadRequest(format!(
+            "Unknown action: {}",
+            action
+        ))),
+    }
+}
+
 pub fn plan_user_delete(is_owner: bool) -> DeletePlan {
     if is_owner {
         DeletePlan::SoftDelete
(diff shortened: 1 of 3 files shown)


src/delete_policy.rs
```diff
@@ -12,6 +12,37 @@ pub enum DeletePlan {
...
+    }
+    Ok(())
+}
+
+/// Maps an action string from a moderation request to a `BlobStatus`.
+/// Returns `Err(BadRequest)` for unknown actions. The DELETE action is
+/// handled separately by callers before reaching this function.
+///
+/// Both `/admin/moderate` and `/admin/api/moderate` accept slightly
+/// different action aliases. This function covers the union so both
```
NotThatKindOfDrLiz: This changes behavior, not just testability. Unifying the alias sets means `/admin/moderate` now accepts `ACTIVE`/`PENDING` and `/admin/api/moderate` now accepts `SAFE`/`QUARANTINE`/`PERMANENT_BAN`. Issue #87 asked for coverage around the current behavior; I do not think we should broaden both endpoint contracts in the same PR.

NotThatKindOfDrLiz (changes requested):
Thanks for doing the extraction work here. Pulling the SHA validation into a shared pure helper is good, and I like having unit coverage around that logic.

I’m requesting changes because `map_moderate_action()` is not a pure refactor: it changes the runtime contract of both moderation endpoints.

## Blocking concerns

1. The shared mapper broadens the accepted alias set for both ingress paths.
   - `/admin/moderate` now accepts `ACTIVE` and `PENDING`, which it did not before.
   - `/admin/api/moderate` now accepts `SAFE`, `QUARANTINE`, and `PERMANENT_BAN`, which it did not before.

   That is a behavior change, not just a testability improvement. Issue #87 asked for coverage around the current behavior, and I do not think we should broaden both endpoint contracts inside that coverage PR.

2. The shared mapper also changes the user-visible error contract for unknown actions.
   - `/admin/moderate` previously returned a more specific error message listing the expected actions.
   - With the shared helper, both endpoints now collapse to the generic `Unknown action: ...`.

## What I’d like to see

- Keep the pure-function extraction, but preserve the current per-endpoint behavior.
- That likely means either:
  - two endpoint-specific pure mappers, or
  - one shared helper parameterized by ingress so each caller keeps its existing alias set and error text.

Separately, this should probably be `Refs #87` rather than `Closes #87` since the missing-blob 404 coverage remains explicitly out of scope here.

src/delete_policy.rs
```diff
@@ -12,6 +12,37 @@ pub enum DeletePlan {
...
+/// endpoints can share validation.
+pub fn map_moderate_action(action: &str) -> Result<BlobStatus> {
+    match action.to_uppercase().as_str() {
+        "BAN" | "BLOCK" | "PERMANENT_BAN" => Ok(BlobStatus::Banned),
+        "RESTRICT" | "QUARANTINE" => Ok(BlobStatus::Restricted),
+        "AGE_RESTRICT" | "AGE_RESTRICTED" => Ok(BlobStatus::AgeRestricted),
+        "APPROVE" | "ACTIVE" | "SAFE" => Ok(BlobStatus::Active),
+        "PENDING" => Ok(BlobStatus::Pending),
+        _ => Err(BlossomError::BadRequest(format!(
```
NotThatKindOfDrLiz, line 35: This also changes the unknown-action error contract. The webhook path used to return a more specific message listing the expected actions; with the shared helper both endpoints now collapse to a generic Unknown action message. If we share this logic, I think we need to preserve each ingress paths existing alias set and error text rather than normalize them.

mbradley:
@NotThatKindOfDrLiz Good catch on both points. Pushed 67826ef which:

1. **Splits the shared mapper into two per-endpoint functions**: `map_admin_api_action()` for `/admin/api/moderate` and `map_webhook_moderate_action()` for `/admin/moderate`. Each preserves its original alias set exactly as it was before this PR.

2. **Restores the webhook's specific error message**: `/admin/moderate` returns `"Unknown action: {action}. Expected BLOCK, RESTRICT, QUARANTINE, AGE_RESTRICTED, or APPROVE"` again.

Tests now pin the boundary explicitly with negative assertions -- e.g., `admin_api_does_not_accept_quarantine`, `webhook_does_not_accept_pending` -- so any future alias-set drift will fail a test.

NotThatKindOfDrLiz (approved):
Approved.

I rechecked the rebased head and the substantive blocker is resolved: the extracted helpers preserve the existing per-endpoint action aliases and unknown-action behavior, and the SHA-256 validation extraction is behavior-equivalent at the converted call sites.

The PR body now matches the implementation, `Refs #87` is correct, and CI is green on the current head. Remaining native test execution work is already tracked in #104, so I'm good to merge this as-is.

PR #105, opened 2026-04-28
refactor(core): extract a blossom-core lib crate for native test execution

### LINKED ISSUE #104: refactor(core): extract a blossom-core lib crate for native test execution
## Problem

All `#[test]` functions in `delete_policy.rs`, `blossom.rs`, and `error.rs` are pure Rust with no Fastly SDK dependency, but `cargo test` fails because the linker can't resolve Fastly's host-call FFI symbols on macOS/Linux. There is no `fastly compute test` subcommand (checked CLI v14.3.1). Tests are currently compile-verified only via `cargo check --tests --target wasm32-wasip1`.

## Proposal

Create a `blossom-core` workspace member (lib crate) containing:
- `BlossomError` / `Result` (replace `fastly::http::StatusCode` with `http::StatusCode` -- same type, both http v1.4.0)
- `BlobStatus`, `BlobMetadata`, and related domain types from `blossom.rs` (zero Fastly imports)
- Pure policy functions from `delete_policy.rs` (`validate_sha256_format`, `map_moderate_action`, `handle_creator_delete_with_ops`, `build_creator_delete_response`, etc.)
- All associated tests (~80 across three modules)

The main binary crate re-exports everything via `pub use blossom_core::*`, so existing `crate::blossom::*` and `crate::error::*` import paths keep working. Fastly-dependent functions (`soft_delete_blob`, `restore_soft_deleted_blob`, `DefaultCreatorDeleteOps`) stay in the binary crate.

## Outcome

`cargo test -p blossom-core` runs all pure-logic tests natively. No Viceroy, no WASM target, standard `cargo test` output with filtering.

## Dependencies

Depends on #103 landing first (adds the `validate_sha256_format` / `map_moderate_action` extraction + tests).

mbradley:
## Summary

- Create `blossom-core` as a workspace library crate for pure Rust domain types and policy helpers with no Fastly SDK dependency
- Move `BlossomError`/`Result`, `BlobStatus`, `BlobMetadata`, related domain types, and pure delete-policy helpers into `blossom-core`
- Preserve the current post-#103 moderation contract by moving the endpoint-specific mappers (`map_admin_api_action`, `map_webhook_moderate_action`) and their negative-assertion tests into `blossom-core`
- Keep Fastly-dependent functions (`soft_delete_blob`, `restore_soft_deleted_blob`, `DefaultCreatorDeleteOps`, `handle_creator_delete`) in the binary crate behind thin shims and re-exports
- Add `cargo test -p blossom-core --locked` to CI so the extracted pure tests actually execute in automation

`cargo test -p blossom-core` now runs **99 tests natively** without Viceroy, and the edge binary still compile-checks against `wasm32-wasip1`.

## Test plan

- [x] `cargo test -p blossom-core --locked` -- 99 passed, 0 failed
- [x] `cargo check --tests --locked` -- compiles clean (pre-existing warnings only)
- [x] CI workflow updated to run `cargo test -p blossom-core --locked`
- [x] Existing binary-crate import paths remain unchanged via `pub use blossom_core::*` shims

Closes #104


NotThatKindOfDrLiz (approved):
Approved.

This is now doing the right thing end to end: the pure types and policy helpers move into `blossom-core`, the current post-#103 moderation contract is preserved there, and CI now actually runs the native `blossom-core` test suite instead of only compile-checking it.

The binary crate stays as the Fastly-facing shim, existing import paths remain stable, and the workspace exclusion keeps the existing `cloud-run-upload` test flow intact. CI is green on the current head, so I'm good to merge.

PR #128, opened 2026-05-08
feat(process-blob): extract  thumbnails for video moderation

### LINKED ISSUE #125: feat: Implement actual thumbnail extraction
### Summary

`extract_video_thumbnail` in `cloud-functions/process-blob/main.py` is a stub: it logs a TODO and always returns `None`. Video uploads therefore never get a real thumbnail for moderation or metadata until this is implemented.

### Problem / Motivation

  `extract_video_thumbnail` in `cloud-functions/process-blob/main.py` is still a
   stub: it logs a TODO and always returns `None`. For `video/*` uploads that means:
  - No JPEG frame is produced or uploaded for moderation.
  - The pipeline falls through to `update_metadata(..., status='pending', ...)` instead of
  running Vision SafeSearch on a thumbnail.
  - Downstream UX (thumbnails at the edge, metadata `thumbnail` field, deleting a thumbnail
   when content is restricted) cannot behave as intended until real extraction exists.

### Proposed Approach

  1. Implement frame extraction from the source video (e.g. `ffmpeg` or equivalent),
  writing a single JPEG to a temp file, then uploading it to the same GCS bucket as the
  source object.
  2. Return the resulting GCS object key from `extract_video_thumbnail` so existing callers
   stay unchanged: `check_image_safety(bucket_name, thumbnail_path)` already uses
  `gs://{bucket}/{blob_name}` for Vision.
  3. Align the **uploaded object key** with how the app serves thumbnails (`{sha256}.jpg`
  vs the current stub `thumbnails/{blob_name}`) so Fastly/Rust (`parse_thumbnail_path`,
  `download_thumbnail`) and the metadata webhook stay consistent.
  4. Extend the `Dockerfile` or chosen runtime
   to include whatever binary/library is required; today only `c2patool` is installed—no
  `ffmpeg` yet.
  5. Ensure extracted `.jpg` objects are ignored by `_is_generated_derivative` / naming
  rules so processing does not recurse on generated thumbnails.
  6. Add or adjust tests around `extract_video_thumbnail` success/failure and the video
  branch of `process_blob_event`.

### Additional Context

  - TODO location: `extract_video_thumbnail`
   https://github.com/divinevideo/divine-blossom/blob/55caf2c5774548f01ed30ed795d384f7d82e976a/cloud-functions/process-blob/main.py#L416
  - Video flow: `process_blob_event` calls extraction, then moderation +
  `handle_moderation_result` with optional `thumbnail_path` for cleanup on flag.
  - Operational limits: gunicorn `--timeout 300` in the Dockerfile; large videos may need
  download/size/time guards analogous to C2PA limits.

sergey3bv:
## Summary
- Implemented real video thumbnail extraction in cloud-functions/process-blob/main.py using ffmpeg to capture the first I frame and upload it as {sha256}.jpg.
- Aligned derivative filtering to canonical generated artifact paths ({sha256}.jpg, {sha256}/hls/..., {sha256}/vtt/...) to avoid over-broad .jpg skipping.
- Expanded cloud-functions/process-blob/test_process_blob.py to cover extraction success/failure paths, video moderation flow behavior, and derivative filtering; fixed flagged-content test to reflect actual deletion behavior.
- Added ffmpeg to cloud-functions/process-blob/Dockerfile runtime dependencies.

## Motivation

- extract_video_thumbnail was a stub returning None, so video uploads could not produce moderation thumbnails and would fall back to pending.
- This change enables the intended moderation path for video uploads and keeps thumbnail storage consistent with existing edge-serving expectations.

## Related Issue

- Closes #125

## Validation

- [ ] `cargo check --tests --locked`
- [ ] `cargo test --manifest-path cloud-run-upload/Cargo.toml --locked`
- [ ] `cargo clippy --locked --all-targets --all-features`

## Manual Test Plan / Notes

- Ran python -m unittest cloud-functions/process-blob/test_process_blob.py (pass, 10 tests).
- Attempted python cloud-functions/process-blob/test/test_c2pa.py, but local run failed due missing flask in environment (ModuleNotFoundError: No module named 'flask').
- Did not run Rust cargo checks because this PR only touches cloud-functions/process-blob Python service and its Dockerfile/tests.

## Visuals / API Examples

- [x] No visual change

## Public Artifact Review


rabble:
**[real-bug]** The diff implements actual thumbnail extraction with ffmpeg, replacing a TODO stub. However, the test mocking setup in `ThumbnailExtractionTests.test_extract_video_thumbnail_success` has a critical bug: it patches `self.module.subprocess.run` but the actual `main.py` imports subprocess at the module level (`import subprocess`), so the patch won't reach the actual call in `extract_video_thumbnail`. This will cause the test to fail when it tries to execute ffmpeg. The patches should target the subprocess module where it's used in main.py, not as a module attribute.

<details><summary>signals</summary>

- Diff adds `import subprocess` to test file and patches `self.module.subprocess.run`
- The actual code uses subprocess directly as `subprocess.run()` from module-level import
- Test mock setup doesn't properly intercept subprocess calls in the main module context
- Similar patch pattern used in other test methods would also fail for the same reason
</details>

_— divine-brain pr-classifier_

sergey3bv:
> **[real-bug]** The diff implements actual thumbnail extraction with ffmpeg, replacing a TODO stub. However, the test mocking setup in `ThumbnailExtractionTests.test_extract_video_thumbnail_success` has a critical bug: it patches `self.module.subprocess.run` but the actual `main.py` imports subprocess at the module level (`import subprocess`), so the patch won't reach the actual call in `extract_video_thumbnail`. This will cause the test to fail when it tries to execute ffmpeg. The patches should target the subprocess module where it's used in main.py, not as a module attribute.

Thanks for the catch, @rabble, in this case the patch target is already correct. `main.py` imports subprocess at module level, so `extract_video_thumbnail()` resolves `subprocess.run` via `main.subprocess.run`. The tests patch `self.module.subprocess.run`, which intercepts that exact call path. I also verified `self.module.subprocess` is subprocess at runtime.

rabble:
this already exists

sergey3bv:
> this already exists

Ok, I will close the issue then

PR #129, opened 2026-05-08
fix(auth): broaden BUD-01 GET acceptance per Postel's law

rabble:
## Summary

Three small relaxations on the BUD-01 GET auth path so the server accepts what real-world clients actually send while still rejecting obviously bad auth events. Follow-up to #127 (which fixed the diagnostic-path dispatch bug for divine-web).

1. **Auth-scheme prefix is now case-insensitive** — \`Nostr \`, \`nostr \`, \`NOSTR \` all parse. RFC 7235 says auth-scheme tokens are case-insensitive; clients have shipped lowercase variants.
2. **The \`t\` action tag is matched case-insensitively** in \`BlossomAuthEvent::get_action()\`, so \`Get\` / \`GET\` / \`gEt\` all map to \`AuthAction::Get\`. The Blossom spec doesn't mandate case.
3. **\`validate_blossom_get_event\` now accepts events with multiple \`x\` tags.** BUD-01 allows authorising multiple blobs in one event, but we previously only checked the first \`x\` tag. Now we iterate every \`x\` tag and accept if any matches the requested hash.

## What's still strict

- Missing \`x\` tag entirely on a kind-24242 GET event — still rejected with a clear "Missing x tag" message. The spec text *does* require it for blob retrieval; better to give clients a clear error than silently accept anonymous-equivalent auth.
- Wrong action (\`t=list\` on a blob path) — still rejected.
- Stale \`expiration\` / \`created_at\` — still rejected.
- Signature / pubkey / event ID validation — unchanged.

## Out of scope (deferred)

Two more Postel relaxations I considered but defer until we see a concrete client hitting them:
- Case-insensitive NIP-98 method tag (\`get\` vs \`GET\`).
- NIP-98 URL canonicalisation (strip trailing slash, ignore query/fragment).

## Test plan

- [x] \`cargo test --lib viewer_auth\` — 30 pass (5 new)
- [x] \`cargo test -p blossom-core\` — 99 pass
- [ ] After merge: deploy via \`fastly compute publish --service-id pOvEEWykEbpnylqst1KTrR\` + purge.

### DIFF
--- a/blossom-core/src/types.rs
+++ b/blossom-core/src/types.rs
@@ -438,11 +438,14 @@ pub enum AuthAction {
 }
 
 impl BlossomAuthEvent {
-    /// Get the action type from tags
+    /// Get the action type from tags. Action value is matched
+    /// case-insensitively per Postel's law — some clients have shipped
+    /// `Get`, `GET`, etc. and the spec doesn't mandate case.
     pub fn get_action(&self) -> Option<AuthAction> {
         for tag in &self.tags {
             if tag.len() >= 2 && tag[0] == "t" {
-                return match tag[1].as_str() {
+                let action = tag[1].to_ascii_lowercase();
+                return match action.as_str() {
                     "get" => Some(AuthAction::Get),
                     "upload" => Some(AuthAction::Upload),
                     "delete" => Some(AuthAction::Delete),
--- a/src/viewer_auth.rs
+++ b/src/viewer_auth.rs
@@ -39,9 +39,20 @@ pub struct ViewerAuthDiagnostics {
 }
 
 pub fn parse_auth_header(auth_header: &str) -> Result<BlossomAuthEvent> {
-    let base64_event = auth_header.strip_prefix("Nostr ").ok_or_else(|| {
+    // Postel's law: accept any case for the scheme name (e.g. `Nostr `,
+    // `nostr `, `NOSTR `). RFC 7235 says auth-scheme is case-insensitive,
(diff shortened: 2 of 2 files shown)


rabble:
**[real-bug]** This PR contains a logic error in the `blossom_get_event_still_rejects_when_no_x_tag_matches` test. The test expects a failure when no x tag matches, but the test setup includes `other_hash` which is used as an x tag value. However, the test then validates against `TEST_HASH`, which will correctly fail. The actual issue is that the test is correctly structured but the test will pass because the validation logic properly iterates through all x tags and rejects only when none match. The code changes themselves (Postel's law relaxations for case-insensitive matching and multi-x tag support) appear sound, but there's a subtle issue: when reporting the hash mismatch error with `last_seen_hash.unwrap_or("")`, the error message quality degrades when there are multiple x tags—it only reports the last one seen rather than all candidates or a more helpful message. Consider improving the error reporting in the hash mismatch case.

<details><summary>signals</summary>

- 
  - The diff adds case-insensitive handling for auth schemes and action tags (per RFC 7235 and Postel's law)
  - The logic change to iterate through all x tags is correct and improves spec compliance (BUD-01)
  - Four new tests are added that validate the relaxations: lowercase scheme, uppercase action, multiple x tags, and rejection cases
  - The error message construction at line 1025 uses `last_seen_hash.unwrap_or("")` which only reports the final hash, not helpful for multi-x scenarios

</details>

_— divine-brain pr-classifier_

PR #132, opened 2026-05-17
fix(asr): Parakeet timestamps fallback + transcoder maxScale 10→20

### LINKED ISSUE #136: Fix Parakeet timestamp crashes and derivative trigger CI flake
## Problem

Parakeet timestamp generation can crash inside NeMo on silent, non-speech, music-only, or very short audio when transcription is requested with `timestamps=True`. During the May 2026 VTT repair/backfill run, these slow failures returned 500s after holding worker capacity long enough to saturate the transcoder pool and cause `/transcode` 429s for unrelated uploads.

At the same time, main CI is red because `cloud-run-upload` derivative trigger tests use a fixed sleep to wait for fire-and-forget localhost requests. Under CI load, the tests verify the mock server before spawned tasks reach it.

## Expected behavior

- Non-speech or degenerate audio should not make the ASR service return a terminal 500 solely because word timestamps could not be computed.
- The transcoder deployment script should preserve the higher max instance cap already used to absorb transcription bursts.
- Upload service tests should wait on observable mock-server activity instead of relying on a fixed short sleep.
- Main CI should return to green.

## Acceptance criteria

- Parakeet retries `transcribe(..., timestamps=False)` when the timestamped path throws.
- The fallback path is covered by a unit test.
- `cloud-run-transcoder/deploy.sh` keeps `--max-instances 20`.
- `cloud-run-upload` derivative trigger tests use bounded condition polling and fail loudly if expected requests never arrive.
- PR checks pass after rebasing on current `main`.

## Validation

- `cd cloud-run-asr-parakeet && python3 -m unittest discover tests`

## Follow-up after merge

- Redeploy `cloud-run-asr-parakeet`.
- Confirm `/v1/transcribe` 500s drop for non-speech inputs.
- Confirm live `divine-transcoder` max scale remains 20 and `/transcode` 429s stay at 0 under scanner/backfill load.

rabble:
## Summary

- Wrap Parakeet `transcribe(..., timestamps=True)` with a fallback retry using `timestamps=False` when NeMo crashes on degenerate timestamp generation.
- Add regression coverage for the timestamped-to-untimed fallback path.
- Persist the live `divine-transcoder` headroom setting by raising `cloud-run-transcoder/deploy.sh` from `--max-instances 10` to `20`.
- Replace flaky fixed-sleep upload derivative trigger tests with bounded polling on `server.received_requests()`.
- Tighten the derivative trigger test helper so it fails loudly on timeout and asserts both expected error-path requests.

## Motivation

Closes #136.

During the May 2026 VTT repair/backfill run, non-speech and other degenerate audio could crash NeMo's timestamp computation after a slow `timestamps=True` call. Those failures returned ASR 500s, held worker capacity, and contributed to transcoder saturation and `/transcode` 429s for unrelated uploads.

Separately, main CI is red because `cloud-run-upload` tests verify WireMock expectations before spawned fire-and-forget derivative tasks reliably reach the local mock server under CI load. The test change removes that timing race without changing production upload behavior.

## Behavior

The ASR service still prefers word timestamps. If that path throws, it logs the timestamp failure and retries the same temp audio file without timestamps, allowing downstream code to return available text or an empty transcript result instead of surfacing a terminal 500 solely because timestamp generation failed.

The upload-service changes are test-only. Tests now wait for the expected number of mock-server requests and panic with the observed count if the deadline expires.

No UI or visual change.

## Manual Validation

- [x] `cd cloud-run-asr-parakeet && python3 -m unittest discover tests` - 7/7 passed
- [x] `cargo test --manifest-path cloud-run-upload/Cargo.toml --locked` - 18/18 passed
- [x] CI green after rebased push

## Post-Merge Follow-Up

- Redeploy `cloud-run-asr-parakeet`.
- Confirm `/v1/transcribe` 500 rate drops for non-speech or degenerate audio inputs.
- Confirm live `divine-transcoder` max scale remains 20 and `/transcode` 429s stay at 0 under scanner/backfill load.


NotThatKindOfDrLiz:
Update pushed after final review:

- Rebasing the branch onto current `main` was clean.
- Added a small follow-up commit, `test(upload): tighten derivative trigger assertions`.
- The upload test polling helper now fails loudly on timeout instead of returning silently.
- The derivative trigger 500-path test now asserts the expected `/transcribe` request as well as `/transcode`.
- Created and linked #136; this PR now closes that issue on merge.

Local validation on the rebased branch:

- `cd cloud-run-asr-parakeet && python3 -m unittest discover tests` - 7/7 passed
- `cargo test --manifest-path cloud-run-upload/Cargo.toml --locked` - 18/18 passed

I am waiting for the refreshed PR checks before approval.

NotThatKindOfDrLiz (approved):
Approved after final pass. The branch is rebased on current main, all refreshed checks are green, #136 is linked with a closing reference, and the added upload-test cleanup removes the silent-timeout assertion gap without changing production behavior.

PR #133, opened 2026-05-24
ci: reduce GitHub Actions runner size

rabble:
## Summary
- replace `ubuntu-24.04-16core` GitHub larger runner labels with `ubuntu-24.04-4core`
- reduce Actions spend while keeping larger-than-standard Linux runner capacity

## Motivation
The Divine org has used 90% of its May 2026 Actions budget. 4-core Linux larger runners are materially cheaper than 16-core runners while still giving build/test jobs more CPU than the default runner.

## Linked issue
None.

## Manual validation
- Verified changed workflow files no longer contain `ubuntu-24.04-16core`.
- Verified changed workflow files now use `ubuntu-24.04-4core`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ env:
 
 jobs:
   test:
-    runs-on: ubuntu-24.04-16core
+    runs-on: ubuntu-24.04-4core
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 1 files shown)


NotThatKindOfDrLiz:
Closing this as not safe/relevant to merge as written.

The original motivation was May 2026 Actions budget pressure, but that premise is now stale. More importantly, the PR only CI run on ubuntu-24.04-4core did not actually execute: the test job sat for 24 hours, was cancelled, had no steps, runner_id 0, and no runner name/group. Current CI on ubuntu-24.04-16core is acquiring runners and passing, so this downgrade has an operational risk beyond slower builds.

This also blanket-reverts the runner sizing from #130 without fresh cost/performance data. If Actions spend is still a concern, the next PR should verify available runner labels first and target measured bottlenecks specifically, rather than changing all compile/build/deploy jobs at once.

PR #135, opened 2026-06-15
fix(transcoder): drop instruction echoes from captions

### LINKED ISSUE #134: bug: Gemini STT echoes its own output-format/schema instructions into captions (prose-instruction leak bypasses the quality gate)
## Summary

Gemini-as-STT in `cloud-run-transcoder` occasionally emits its own output-format / JSON-schema **instructions** as transcript text. Because the leaked text is plain prose (no JSON braces, no repeated loop), it passes every deterministic quality-gate heuristic and ships as a caption.

Reported downstream as divinevideo/divine-mobile#3737. Example caption (real speech first, echoed instructions second):

> "…Well, that's not really freedom now, is it, you freaking idiot? **a single JSON array. Do not include any extra text outside of the JSON string. When producing JSON you must follow the schema provided in the context.**"

The reported video (`sha256 4ad97fcf5fee46f7cf4b794b355e6fb0ee08c0e35d80098b926cf2441ba582e3`) has already self-healed — `media.divine.video/<sha256>/vtt` now returns a clean transcript — but the generation gap remains and will recur.

## Root cause

- `build_gemini_prompt()` (`src/main.rs:2654-2688`) sends the strict JSON/`responseSchema` output-format directives in the **same user turn** as the audio (request built `src/main.rs:2711-2740`). Gemini sometimes restates those directives inside a `segments[].text` field.
- `normalize_transcript_to_vtt` → `strip_envelope_explanation_tail` (`src/main.rs:3044`) is a no-op on bare prose — it only cuts at ``` fences, `{`/`[` structural starts, `<response`, or a fixed preamble list. "JSON"-as-prose matches none of these, so the echo is written verbatim into the cue body.

## Why the quality gate misses it

Every deterministic check keys on JSON **syntax**, repetition, density, or silence:

- `transcript_drop_reason` (`src/main.rs:3245`, run for all providers) → `is_implausible_text_density` (>40 cps), `is_loop_hallucination` (verbatim probe repeats ≥3×), `should_drop_low_signal_transcript` (low-signal audio + URL/outro/courtesy phrase), low-confidence (Gemini supplies no logprobs). None fire for normal-speech + a single non-repeating instruction sentence.
- `google_drop_reason` (`src/transcription_google_stt_v2.rs:753`) — which has `contains_provider_json_artifact` / `is_repeated_phrase_hallucination` / `is_non_speech_garbage` — is **provider-gated to `google_stt_v2`** (`src/main.rs:1805`) and never runs on the Gemini path.
- The Python rescanners (`scan_and_repair_vtts.py:classify_vtt`, `repair_recent_bad_vtts.py:is_bad_vtt_body`) key on the same JSON markers, so they also miss it. Only the optional, flag-gated LLM judge (`scan_and_repair_vtts.py`) would flag "model meta-commentary."

Related prior art: #38 hardened the JSON-envelope path; this is the prose-instruction variant that escaped that hardening.

## Proposed fix

- **Prevention:** forbid restating instructions/schema in the prompt, and move the format directives into Vertex `system_instruction` (top-level on the request body) rather than the audio-carrying user turn. Reduces but won't eliminate echo.
- **Detection (primary):** add a `contains_instruction_echo(text)` guard wired into `transcript_drop_reason` (new `TranscriptDropReason` variant) so it runs for **all** providers — flag transcripts containing ≥2 telltale output-format phrases vanishingly rare in speech (e.g. `json array`, `json object`, `json string`, `follow the schema`, `schema provided`, `extra text outside`, `output requirements`, `valid json`, `bcp47`, `<seconds>`, `spoken words`). Mirror the same list into `scan_and_repair_vtts.py:classify_vtt` and `repair_recent_bad_vtts.py:is_bad_vtt_body`.
- **Repair:** rescan stored VTTs with the new classifier and retranscribe matches; keep the optional LLM judge enabled as a catch-all.

## Out of scope

No divine-mobile change — the app fetches and renders first-party VTT verbatim, which is the correct contract.

NotThatKindOfDrLiz:
## Summary
- add a universal transcoder quality-gate reason for leaked output-format/schema instruction echoes in captions
- mirror the detector into both VTT repair scanners so stored bad captions can be found and retranscribed
- add Rust and Python regression coverage for the real failure shape and technical-speech false-positive boundaries

## Motivation
Gemini-as-STT can echo its own JSON/schema output instructions into segment text as plain prose. Existing structural JSON cleanup and provider-specific Google guards do not catch that failure mode, so the echoed instructions can ship as WebVTT captions.

Closes #134.

## Root Cause
The Gemini request asks for strict structured transcript output. When Gemini copies output-format/schema prose into a segment text field, the existing cleanup only strips structural JSON/fence tails, and the universal transcript drop gate only checks confidence, text density, loops, and low-signal hallucinations.

## User / Developer Impact
Bad instruction-echo captions are now dropped before upload, which triggers the existing empty-VTT/self-heal behavior rather than serving leaked model instructions. The operational scanners now classify the same prose-instruction leak as `instruction_echo` for repair runs. The detector requires prompt/schema-specific phrase co-occurrence to avoid dropping normal developer or API-related speech.

## Manual Validation
- `python3 -m unittest cloud-run-transcoder/tests/test_repair_recent_bad_vtts.py cloud-run-transcoder/tests/test_scan_and_repair_vtts.py`
- `python3 -m py_compile cloud-run-transcoder/repair_recent_bad_vtts.py cloud-run-transcoder/scan_and_repair_vtts.py cloud-run-transcoder/tests/test_repair_recent_bad_vtts.py cloud-run-transcoder/tests/test_scan_and_repair_vtts.py`

## Notes
No visual change. Root-level clippy/check commands exit successfully but still print pre-existing warnings in unrelated Fastly/root code.

### DIFF
--- a/cloud-run-transcoder/repair_recent_bad_vtts.py
+++ b/cloud-run-transcoder/repair_recent_bad_vtts.py
@@ -22,6 +22,33 @@ JSON_CORRUPTION_MARKERS = (
     '"total_tokens"',
     '"usage":{',
     '"prompt_tokens"',
+    '"completion_tokens"',
+    '"finish_reason"',
+)
+INSTRUCTION_ECHO_STRONG_MARKERS = (
+    "<bcp47>",
+    "<seconds>",
+    "<spoken words>",
+    "output requirements",
+    "follow the schema",
+    "provided in the context",
+    "extra text outside",
+    "outside of the json",
+    "do not include any extra text",
+    "return only a json",
+    "only a json object",
+    "single json array",
+    "spoken words transcribed verbatim",
+    "text field of every segment",
+)
+INSTRUCTION_ECHO_WEAK_MARKERS = (
+    "valid json",
+    "json array",
+    "json object",
+    "json string",
+    "markdown",
+    "code fences",
+    "response schema",
 )
 USER_AGENT = "divine-blossom/repair_recent_bad_vtts"
 
@@ -85,7 +112,25 @@ def collect_recent_media_hashes(
 
 
 def is_bad_vtt_body(body: str) -> bool:
+    return has_json_artifact(body) or has_instruction_echo(body)
+
+
+def has_json_artifact(body: str) -> bool:
     return any(marker in body for marker in JSON_CORRUPTION_MARKERS)
+
+
+def has_instruction_echo(text: str) -> bool:
+    """Mirror of cloud-run-transcoder/src/main.rs `contains_instruction_echo`."""
+    normalized = " ".join(text.split()).lower()
+    strong_count = sum(
+        1 for marker in INSTRUCTION_ECHO_STRONG_MARKERS if marker in normalized
+    )
+    if strong_count >= 2:
+        return True
+    weak_count = sum(
+        1 for marker in INSTRUCTION_ECHO_WEAK_MARKERS if marker in normalized
+    )
+    return strong_count >= 1 and weak_count >= 2
 
 
 def is_empty_vtt_body(body: str) -> bool:
--- a/cloud-run-transcoder/repair_recent_bad_vtts.py
+++ b/cloud-run-transcoder/repair_recent_bad_vtts.py
@@ -11,45 +11,25 @@ import sys
 import time
 from dataclasses import dataclass
 from datetime import datetime, timedelta, timezone
+from pathlib import Path
 from typing import Any
 from urllib import error, parse, request
+
+# Share one definition of the quality heuristics with the scanner so the two
+# operator tools (and their marker lists) cannot drift apart. The scan module
+# is a sibling script with no import-time side effects.
+sys.path.insert(0, str(Path(__file__).resolve().parent))
+from scan_and_repair_vtts import (  # noqa: E402
+    has_instruction_echo,
+    has_json_artifact,
+    has_json_envelope_leak,
+    vtt_spoken_text,
+)
 
 
 RELAY_API = "https://relay.divine.video/api"
 MEDIA_URL = "https://media.divine.video"
 REQUEST_TIMEOUT = 30
-JSON_CORRUPTION_MARKERS = (
-    '"total_tokens"',
-    '"usage":{',
-    '"prompt_tokens"',
-    '"completion_tokens"',
-    '"finish_reason"',
-)
-INSTRUCTION_ECHO_STRONG_MARKERS = (
-    "<bcp47>",
-    "<seconds>",
-    "<spoken words>",
-    "output requirements",
-    "follow the schema",
-    "provided in the context",
-    "extra text outside",
-    "outside of the json",
-    "do not include any extra text",
-    "return only a json",
-    "only a json object",
-    "single json array",
-    "spoken words transcribed verbatim",
-    "text field of every segment",
-)
-INSTRUCTION_ECHO_WEAK_MARKERS = (
-    "valid json",
-    "json array",
-    "json object",
-    "json string",
-    "markdown",
-    "code fences",
-    "response schema",
-)
 USER_AGENT = "divine-blossom/repair_recent_bad_vtts"
 
 
@@ -111,26 +91,26 @@ def collect_recent_media_hashes(
     return hashes
 
 
+def bad_vtt_reason(body: str) -> str | None:
+    """Classify a VTT body the same way the deployed gate and scanner do.
+
+    Detectors run on the *spoken* text (not the raw body) so a leaked
+    instruction split across cue boundaries — the normal one-cue-per-segment
+    shape — is still seen as a contiguous phrase. Returns a short reason
+    string for the operator log, or None when the body looks clean.
+    """
+    if has_json_artifact(body):
+        return "json_artifact"
+    spoken = vtt_spoken_text(body)
+    if has_json_envelope_leak(spoken):
+        return "json_envelope_leak"
+    if has_instruction_echo(spoken):
+        return "instruction_echo"
+    return None
+
+
 def is_bad_vtt_body(body: str) -> bool:
-    return has_json_artifact(body) or has_instruction_echo(body)
-
-
-def has_json_artifact(body: str) -> bool:
-    return any(marker in body for marker in JSON_CORRUPTION_MARKERS)
-
-
-def has_instruction_echo(text: str) -> bool:
-    """Mirror of cloud-run-transcoder/src/main.rs `contains_instruction_echo`."""
-    normalized = " ".join(text.split()).lower()
-    strong_count = sum(
-        1 for marker in INSTRUCTION_ECHO_STRONG_MARKERS if marker in normalized
-    )
-    if strong_count >= 2:
-        return True
-    weak_count = sum(
-        1 for marker in INSTRUCTION_ECHO_WEAK_MARKERS if marker in normalized
-    )
-    return strong_count >= 1 and weak_count >= 2
+    return bad_vtt_reason(body) is not None
 
 
 def is_empty_vtt_body(body: str) -> bool:
@@ -342,16 +322,16 @@ def main() -> int:
                 print(f"SKIP {media_hash}: vtt status={status}")
             continue
 
-        bad_vtt = is_bad_vtt_body(body)
+        reason = bad_vtt_reason(body)
         empty_vtt = args.check_empty and is_empty_vtt_body(body)
-        if not bad_vtt and not empty_vtt:
+        if reason is None and not empty_vtt:
             summary.clean_vtts += 1
             if args.verbose:
                 print(f"CLEAN {media_hash}")
             continue
 
         summary.bad_vtts += 1
-        reason = "json" if bad_vtt else "empty"
+        reason = reason or "empty"
         print(f"BAD {media_hash}: {reason}")
 
         if args.dry_run:
--- a/cloud-run-transcoder/src/main.rs
+++ b/cloud-run-transcoder/src/main.rs
@@ -3294,11 +3294,52 @@ fn transcript_drop_reason(
     None
 }
 
+/// Count distinct instruction-phrase clusters in `normalized`.
+///
+/// Collects every marker match span, then merges spans that overlap *or* are
+/// separated only by whitespace, so one contiguous instruction sentence counts
+/// as a single phrase rather than several. This is what makes the threshold in
+/// `contains_instruction_echo` mean "multiple distinct phrases": redundant,
+/// overlapping markers (e.g. "return only a json" + "only a json object" inside
+/// "return only a JSON object") collapse to one cluster instead of inflating the
+/// count and tripping the gate on ordinary speech.
+fn distinct_marker_phrases(normalized: &str, markers: &[&str]) -> usize {
+    let mut spans: Vec<(usize, usize)> = Vec::new();
+    for marker in markers {
+        let mut from = 0;
+        while let Some(rel) = normalized[from..].find(marker) {
+            let start = from + rel;
+            spans.push((start, start + marker.len()));
+            from = start + 1;
+        }
+    }
+    if spans.is_empty() {
+        return 0;
+    }
+    spans.sort_unstable();
+
+    let mut clusters = 1;
+    let mut cluster_end = spans[0].1;
+    for &(start, end) in &spans[1..] {
+        // Same cluster when overlapping, or separated only by whitespace
+        // (one contiguous instruction). A non-whitespace gap means another
+        // distinct phrase.
+        if start <= cluster_end || normalized[cluster_end..start].trim().is_empty() {
+            cluster_end = cluster_end.max(end);
+        } else {
+            clusters += 1;
+            cluster_end = end;
+        }
+    }
+    clusters
+}
+
 /// Detect model output-format instructions that leaked into caption text.
 ///
-/// This intentionally requires multiple prompt/schema-derived phrases. Common
-/// technical speech such as "valid JSON" or "JSON array" alone must not trip a
-/// universal provider guard.
+/// This intentionally requires multiple *distinct* prompt/schema-derived
+/// phrases (see `distinct_marker_phrases`). Common technical speech such as
+/// "valid JSON" or "JSON array" alone — or a single contiguous instruction
+/// sentence — must not trip a universal provider guard.
 fn contains_instruction_echo(text: &str) -> bool {
     let normalized = text
         .split_whitespace()
@@ -3332,18 +3373,12 @@ fn contains_instruction_echo(text: &str) -> bool {
         "response schema",
     ];
 
-    let strong_count = STRONG_MARKERS
-        .iter()
-        .filter(|marker| normalized.contains(**marker))
-        .count();
+    let strong_count = distinct_marker_phrases(&normalized, STRONG_MARKERS);
     if strong_count >= 2 {
         return true;
     }
 
-    let weak_count = WEAK_MARKERS
-        .iter()
-        .filter(|marker| normalized.contains(**marker))
-        .count();
+    let weak_count = distinct_marker_phrases(&normalized, WEAK_MARKERS);
     strong_count >= 1 && weak_count >= 2
 }
 
@@ -4132,6 +4167,21 @@ mod tests {
         let tech_talk = "In our API docs, follow the schema provided for each endpoint \
             before sending the request body.";
         assert!(!contains_instruction_echo(tech_talk));
+    }
+
+    #[test]
+    fn instruction_echo_guard_passes_overlapping_strong_markers() {
+        // Ordinary speech where several STRONG markers overlap inside one
+        // contiguous clause must not reach the >=2 threshold. Before the
+        // cluster fix this dropped a whole transcript on a single sentence.
+        assert!(!contains_instruction_echo(
+            "The endpoint should return only a JSON object with the user's data."
+        ));
+        // A single contiguous instruction phrase (three overlapping markers)
+        // is one phrase, not three — needs a second distinct phrase to fire.
+        assert!(!contains_instruction_echo(
+            "Do not include any extra text outside of the JSON string."
+        ));
     }
 
     #[test]
--- a/cloud-run-transcoder/src/main.rs
+++ b/cloud-run-transcoder/src/main.rs
@@ -1799,6 +1799,14 @@ async fn process_transcribe(
             );
             ParsedVtt::empty(audio_analysis.duration_ms)
         }
+        Some(TranscriptDropReason::InstructionEcho) => {
+            warn!(
+                hash,
+                text_preview = %first_chars(&parsed_vtt.text, 120),
+                "Dropping transcript with leaked output-format instructions"
+            );
+            ParsedVtt::empty(audio_analysis.duration_ms)
+        }
         None => parsed_vtt,
     };
 
@@ -3240,6 +3248,7 @@ enum TranscriptDropReason {
     LowSignalHeuristic,
     LoopHallucination,
     ImplausibleTextDensity,
+    InstructionEcho,
 }
 
 fn transcript_drop_reason(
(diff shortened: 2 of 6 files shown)


realmeylisdev:
Two edge cases on the marker heuristic, for awareness — not blockers:

- **False positive**: legitimate developer/API speech that hits 2+ strong markers — e.g. "…follow the schema provided in the API docs…" ("follow the schema" + "schema provided") — drops to an empty VTT. There's no semantic distinction from a real echo, so it's an inherent limit of substring matching; flagging that the failure mode is permanent caption loss for that transcode, not a recoverable state.
- **False negative**: rephrased echoes at 1 strong + 1 weak slip through (e.g. "return a JSON object and follow the schema from the context").

The conservative 2-strong-OR-(1-strong+2-weak) threshold keeps both bounded, and "drop → empty → self-heal/retranscribe" is the right failure direction. Might be worth a regression case for the FP boundary — a 2-strong-marker legitimate-speech sample that should NOT drop (current tests cover a 2-weak pass case but not the 2-strong collision).

Minor attribution note: the leaked phrases aren't in `build_gemini_prompt` in any version (grep + git history) — they're Gemini's own responseSchema boilerplate. Doesn't affect this PR's detector, but it means the "move to system_instruction" prevention won't address the root signal.

cloud-run-transcoder/src/main.rs
```diff
@@ -3274,13 +3283,70 @@ fn transcript_drop_reason(
...
+        "markdown",
+        "code fences",
+        "response schema",
+    ];
+
+    let strong_count = STRONG_MARKERS
+        .iter()
+        .filter(|marker| normalized.contains(**marker))
+        .count();
+    if strong_count >= 2 {
```
dcadenas: A single ordinary sentence can trip this gate and wipe a whole transcript.
The `strong_count >= 2` check counts overlapping substrings as separate evidence.
The one phrase "do not include any extra text outside of the JSON" already contains three strong markers at once:

https://github.com/divinevideo/divine-blossom/blob/543e7bb2a991341d7f66d909b44164b42191dfc5/cloud-run-transcoder/src/main.rs#L3316-L3324

So that one sentence fires the gate by itself, which is exactly what the comment at lines 3299-3301 says must not happen.
A developer-tutorial caption that says that sentence, or one strong phrase plus two weak JSON nouns like "follow the schema, return a JSON object, no markdown", gets replaced with an empty VTT and retranscribed.
Count only non-overlapping marker spans, or distinct phrase positions, so a single contiguous instruction sentence cannot reach `strong_count >= 2` on its own.

cloud-run-transcoder/repair_recent_bad_vtts.py
```diff
@@ -85,9 +112,27 @@ def collect_recent_media_hashes(
 
 
 def is_bad_vtt_body(body: str) -> bool:
+    return has_json_artifact(body) or has_instruction_echo(body)
```
dcadenas: This check runs on the raw VTT body, so it can miss the bad captions it was built to find.
`has_instruction_echo(body)` here matches against the full VTT text, which still has the WEBVTT header, cue numbers, and timestamp lines mixed in.
The live gate at main.rs:3286 and the sibling scanner both match spoken text only:

https://github.com/divinevideo/divine-blossom/blob/543e7bb2a991341d7f66d909b44164b42191dfc5/cloud-run-transcoder/scan_and_repair_vtts.py#L444-L448

When a leaked instruction spans more than one cue, a cue index or timestamp line sits between the words. The marker phrase is then no longer a contiguous substring, so the scanner passes it silently.
Run the check on spoken text here too. Better, import `has_instruction_echo`, the spoken-text extractor, and the marker lists from `scan_and_repair_vtts` so both Python tools share one definition and one input.
Right now each file keeps its own copy of the marker lists (lines 28-52 here, 54-78 there), so they will drift again on the next edit.

dcadenas (changes requested):
This adds a useful safeguard against leaked model instructions ending up in captions, and the three-place mirror is reasonable. Two things have to change before merge. The central gate is weaker than its own comment promises and can drop a real transcript from a single sentence. One of the repair tools runs on the wrong input and can miss the captions it exists to find. Details are in the two inline comments.

cloud-run-transcoder/src/main.rs
realmeylisdev: Confirming this over-fires — reproduced independently against the repo's exact marker lists. `strong_count` sums independent `.contains()` checks (`main.rs:3335-3338`) with no de-overlapping, and STRONG_MARKERS contains mutually-overlapping phrases, so a single natural clause hits the `strong_count >= 2` short-circuit:

- `"The endpoint should return only a JSON object with the user's data."` → `return only a json` + `only a json object` = **2 strong → whole transcript dropped**
- `"…follow the schema provided in the context…"` → `follow the schema` + `provided in the context` = 2 strong → dropped
- `"do not include any extra text outside of the JSON string"` → 3 strong (`do not include any extra text`, `extra text outside`, `outside of the json`)

The 1-strong + 2-weak branch (`3343-3347`) also fires on plausible technical speech: `"Follow the schema … a JSON array, a JSON object … response schema."` → 1 strong + 3 weak → dropped.

This contradicts the docstring at `3297-3301`. And the consequence is permanent: on drop the code returns `ParsedVtt::empty` (`:1808`), which is uploaded and reported `status=complete` (`:3460-3474`) and cached at the edge (`divine-blossom/src/main.rs:1605-1626`) with no auto-retranscribe — so a false positive removes captions for that video until someone runs a manual `force=true` repair.

Counting **non-overlapping** marker spans (or distinct phrase positions) would restore the "multiple distinct phrases" contract.

cloud-run-transcoder/repair_recent_bad_vtts.py
realmeylisdev: `has_instruction_echo` here runs on the **raw VTT body**, unlike the live gate (`contains_instruction_echo(&parsed_vtt.text)`, joined spoken text — `main.rs:3286`) and `scan_and_repair_vtts.py` (`text = vtt_spoken_text(body)` then `has_instruction_echo(text)` — `:445-448`).

After `" ".join(text.split())`, the WEBVTT header, cue-index, and `00:00:..-->` timestamp tokens are folded inline. The normal Gemini path emits **one cue per segment** (`main.rs:3062-3069`), so a leaked instruction split across cue boundaries is no longer a contiguous substring and is missed here. Reproduced: a strong marker straddling a cue boundary gives `is_bad_vtt_body(body) → False` while `scan.classify_vtt(body) → "instruction_echo"` for the same bytes — i.e. the repair tool skips a caption the deployed gate would have caught.

This file also has no spoken-text extractor, and (separately) `is_bad_vtt_body` lacks the `json_envelope_leak` branch that `scan` has. Importing `vtt_spoken_text` + `has_instruction_echo` + the marker lists from `scan_and_repair_vtts.py` (or a shared module) would fix the input shape and collapse the duplicated marker lists in one change.

cloud-run-transcoder/tests/test_repair_recent_bad_vtts.py
```diff
@@ -47,6 +47,46 @@ def test_ignores_valid_vtt_body(self):
...
+        detector = getattr(module, "is_bad_vtt_body", None)
+        self.assertIsNotNone(detector, "is_bad_vtt_body should exist")
+
+        body = (
+            "WEBVTT\n\n1\n00:00:00.000 --> 00:00:08.000\n"
+            "In our API docs, follow the schema provided for each endpoint "
+            "before sending the request body.\n"
+        )
+
+        self.assertFalse(detector(body))
```
realmeylisdev, line 88: These negative tests only exercise 0–1 strong markers (`test_ignores_technical_json_speech` has 0 strong; this one has exactly 1, `follow the schema`), so they never reach the `strong_count >= 2` short-circuit or the `1-strong + 2-weak` branch — the two boundaries where the gate actually over-fires.

Recommend adding regression cases that must **not** drop:
- `"The endpoint should return only a JSON object with the user's data."` (2 overlapping strong markers)
- `"Follow the schema … a JSON array, a JSON object … response schema."` (1 strong + 2 weak)
- a multi-cue body where a marker is split across cues (the raw-body path in this file)

Mirror the same in the Rust tests (`main.rs:4115-4166`) and the `scan_and_repair_vtts.py` tests.

realmeylisdev (changes requested):
## Formal review — independent verification

I reproduced every claim below against the source and ran the suites: **Python 9/9** and the **Rust `instruction_echo` tests 5/5** pass, and the transcoder crate builds (only a pre-existing dead-code warning). The approach is sound and well-targeted, the three detector implementations are byte-identical, and the cross-repo blast radius is clean — verified end-to-end:

- **divine-mobile** renders first-party VTT verbatim (`SubtitleService.parseVtt`, no caption transform) and degrades an empty `WEBVTT` body to "no caption" gracefully (`subtitle_overlay.dart` → `SizedBox.shrink()`); the client `TranscriptSanitizer` is **not** on `main` (only the abandoned `fix-issue-3737` branch), so backend stays the single home for this fix.
- **divine-web** (`useSubtitles.ts`) and **divine-moderation-service** (`pipeline.mjs`) both no-op on an empty VTT (the moderation path stops feeding garbage into classify/topics — a net improvement).
- The serving path is **terminal**: an emptied transcript is uploaded and reported `status=complete` and cached at the edge — there is **no** auto-retranscribe loop.

The #134 attribution also checks out: the leaked phrases (`single JSON array`, `extra text outside of the JSON`, `follow the schema provided in the context`) appear in **no** prompt or source file in any repo, and `build_gemini_prompt` never contained them across its full git history (3 versions) — they are Gemini `responseSchema` boilerplate. The marker list correctly covers both that boilerplate and the prompt-derived phrases.

I confirm **both** of @dcadenas's findings (independently reproduced) and add several more. **Requesting changes.**

### Blocking

**1. `contains_instruction_echo` over-fires via overlapping markers** (`cloud-run-transcoder/src/main.rs:3335-3347`). `strong_count` sums independent `.contains()` tests, so overlapping marker substrings in a single clause each add to the count — one ordinary sentence reaches the threshold, contradicting the docstring's "intentionally requires multiple prompt/schema-derived phrases." This drop is **non-recoverable** (empty VTT → `status=complete` → edge-cached, no auto-retranscribe), so any false positive permanently removes captions until a manual `force=true` repair. Details + verified examples inline.

**2. `repair_recent_bad_vtts.py` scans the raw VTT body** (`:115`), unlike the live gate and `scan_and_repair_vtts.py`, which use spoken text. A leaked instruction split across cues (the normal one-cue-per-segment output shape) is missed — the repair tool can skip the captions it exists to find. Details inline.

### Non-blocking

**3. `--check-empty` re-trigger churn** (`repair_recent_bad_vtts.py:136-138`, `345-371`). The guard's `"WEBVTT\n\n"` output matches `is_empty_vtt_body`, so with `--check-empty` every guard-emptied video is re-classified `empty` and force-retranscribed on **every** operator run; if Gemini re-emits the same boilerplate deterministically, the video never converges. Not an unattended loop (operator-invoked, no cron) and the default mode is safe, but worth distinguishing "intentionally emptied by a drop guard" from "never transcribed" (or a per-hash cooldown).

**4. Misleading operator log label** (`repair_recent_bad_vtts.py:354`). `reason = "json" if bad_vtt else "empty"` prints an instruction-echo detection as `BAD <hash>: json`. `scan_and_repair_vtts.py` already emits the precise `instruction_echo` reason.

**5. Marker lists triplicated** (`main.rs:3309-3334`, `scan_and_repair_vtts.py:54-78`, `repair_recent_bad_vtts.py:28-52`) — byte-identical today, but nothing enforces parity. A shared fixture both languages load, or a CI parity check, would prevent drift. (Fixing #2 by sharing one definition also collapses the two Python copies.)

**6. False-positive boundaries are untested** — see inline on the test file.

NotThatKindOfDrLiz:
Addressed the requested changes and pushed `18f86f2`.

Summary:
- Fixed the overlapping-marker false positive by counting distinct strong instruction-phrase clusters instead of raw substring hits.
- Fixed `repair_recent_bad_vtts.py` to classify spoken VTT text, not raw VTT bytes, so instruction echoes split across cues are detected.
- Removed the remaining `1 strong + 2 weak` false-positive path: generic JSON/schema terms no longer contribute to the drop decision, so technical speech like “Follow the schema ... JSON array ... JSON object ... response schema” is preserved.
- Shared the Python detector path between the repair tools, added `json_envelope_leak` parity, and changed repair logs to emit precise reasons such as `instruction_echo`.
- Added regression coverage in Rust and both Python suites for overlapping strong markers, one-strong-plus-generic-JSON speech, and cross-cue instruction echoes.

Validation:
- `python3 -m unittest cloud-run-transcoder/tests/test_repair_recent_bad_vtts.py cloud-run-transcoder/tests/test_scan_and_repair_vtts.py` — 14 passed
- `PYTHONPYCACHEPREFIX=/private/tmp/divine-blossom-pycache python3 -m py_compile cloud-run-transcoder/repair_recent_bad_vtts.py cloud-run-transcoder/scan_and_repair_vtts.py cloud-run-transcoder/tests/test_repair_recent_bad_vtts.py cloud-run-transcoder/tests/test_scan_and_repair_vtts.py` — passed
- `cargo test --manifest-path cloud-run-transcoder/Cargo.toml --locked` — 123 passed
- `cargo clippy --manifest-path cloud-run-transcoder/Cargo.toml --locked --all-targets` — passed with existing unrelated warnings only

I intentionally left the non-blocking `--check-empty` re-trigger churn out of this PR. It is pre-existing operator-tooling behavior behind a non-default flag and should be handled as a separate follow-up rather than broadening this fix.

cloud-run-transcoder/src/main.rs
```diff
@@ -3274,13 +3283,89 @@ fn transcript_drop_reason(
...
+        .collect::<Vec<_>>()
+        .join(" ")
+        .to_ascii_lowercase();
+
+    const STRONG_MARKERS: &[&str] = &[
+        "<bcp47>",
+        "<seconds>",
+        "<spoken words>",
+        "output requirements",
+        "follow the schema",
```
realmeylisdev: Non-blocking follow-up — residual over-fire on ordinary technical speech.

The overlapping-marker inflation from the first review is fixed (cluster counting verified). The residual: `follow the schema` and `provided in the context` are generic English and were created by splitting the actually-leaked phrase `follow the schema provided in the context` in two. Contiguously (the real leak) they merge to one cluster — so they add little to true detection; when ordinary speech puts words between them they become two clusters and fire. Reproduced on this branch (Rust and Python agree):

- `Our API output requirements changed, so please follow the schema in the new docs.` → `output requirements` + `follow the schema` → 2 clusters → dropped
- `You should follow the schema that was provided in the context of the previous lesson.` → `follow the schema` + `provided in the context` → 2 clusters → dropped

Each is a legitimate transcript, and the drop is terminal (empty VTT → status=complete → immutable edge cache, no auto-retranscribe), so the caption is lost until a manual `force=true` repair.

Small, low-risk tightening for a follow-up: collapse `follow the schema` + `provided in the context` into the single contiguous marker `follow the schema provided in the context` (what actually leaked), and/or require at least one high-specificity marker (`<bcp47>`, `<seconds>`, `<spoken words>`, `spoken words transcribed verbatim`, `text field of every segment`, the JSON-object phrases) before firing. The real #134 leak carries `single json array` / `do not include any extra text` / `outside of the json`, so it still fires; the false positives above stop.

cloud-run-transcoder/src/main.rs
```diff
@@ -3274,13 +3283,89 @@ fn transcript_drop_reason(
...
+    }
+    spans.sort_unstable();
+
+    let mut clusters = 1;
+    let mut cluster_end = spans[0].1;
+    for &(start, end) in &spans[1..] {
+        // Same cluster when overlapping, or separated only by whitespace
+        // (one contiguous instruction). A non-whitespace gap means another
+        // distinct phrase.
+        if start <= cluster_end || normalized[cluster_end..start].trim().is_empty() {
```
realmeylisdev, line 3327: Non-blocking — Rust/Python parity gap on C0 control chars. `normalized[cluster_end..start].trim()` (Rust) and `normalized[cluster_end:start].strip()` (Python) disagree for U+001C–U+001F: Python's `split()`/`strip()` treat them as whitespace, Rust's `split_whitespace()`/`trim()` (Unicode White_Space) do not. So `output requirements<FS>follow the schema` is 2 clusters (drop) in Rust but 1 (keep) in Python — the live gate and scanner would disagree. Low practical impact (these bytes rarely reach VTT spoken text), but it breaks the byte-for-byte parity. Normalizing C0 controls to a space in both closes it.

cloud-run-transcoder/tests/test_scan_and_repair_vtts.py
```diff
@@ -0,0 +1,94 @@
+import importlib.util
```
realmeylisdev, line 1: Non-blocking — these tests don't run in CI. `cloud-run-transcoder` is excluded from the root Cargo workspace (`Cargo.toml`: `exclude = ["cloud-run-upload", "cloud-run-transcoder"]`) and `ci.yml` only runs `blossom-core` + `cloud-run-upload` cargo tests + clippy — no `cargo test --manifest-path cloud-run-transcoder/...` and no Python step. So the new Rust `instruction_echo` tests and both Python suites here never execute in CI, and nothing guards the three-way marker/logic parity (prior non-blocking #5). A CI step running the transcoder cargo tests + `python -m unittest` for these suites + a marker-list parity assertion (load the Rust list and the Python list, assert equal) would make this PR's regression protection real.

cloud-run-transcoder/repair_recent_bad_vtts.py
```diff
@@ -84,8 +91,26 @@ def collect_recent_media_hashes(
...
+    """Classify a VTT body the same way the deployed gate and scanner do.
+
+    Detectors run on the *spoken* text (not the raw body) so a leaked
+    instruction split across cue boundaries — the normal one-cue-per-segment
+    shape — is still seen as a contiguous phrase. Returns a short reason
+    string for the operator log, or None when the body looks clean.
+    """
+    if has_json_artifact(body):
+        return "json_artifact"
+    spoken = vtt_spoken_text(body)
```
realmeylisdev, line 104: Confirmed fixed — blocking #2 from the prior review. `bad_vtt_reason` now classifies `vtt_spoken_text(body)` rather than the raw body, so a leaked instruction split across the normal one-cue-per-segment boundaries is seen as contiguous and detected (verified via `test_detects_instruction_echo_split_across_cues` and end-to-end). Ordering matches the scanner's `classify_vtt` (json_artifact on the raw body, envelope/echo on spoken text).

realmeylisdev (approved):
## Formal re-review — Approve

Re-reviewed `18f86f2` (the `cdf7716` + `18f86f2` follow-ups to the prior CHANGES_REQUESTED). I re-ran the suites and traced the live gate, the operator scanners, and the full downstream blast radius across repos. Both prior blocking items are resolved and the cross-repo impact is clean — approving. Non-blocking notes below (one worth a follow-up).

**Tests (local):** transcoder Rust 123/123; the 7 `instruction_echo` tests pass; both Python suites 14/14. Rust and Python marker lists are byte-identical (14 markers).

### Prior blocking items — both resolved
- **#1 (over-fire via overlapping markers):** Resolved. `distinct_marker_phrases` merges overlapping/whitespace-adjacent spans into clusters; `"return only a JSON object with the user's data"` and a single contiguous instruction sentence both stay at 1 cluster and no longer drop. Removing the weak markers makes the gate strictly more conservative.
- **#2 (repair scanned the raw body):** Resolved. `repair_recent_bad_vtts.py` now runs the detector on `vtt_spoken_text(body)`, matching the live gate and the scanner; a cross-cue split echo is detected. Sharing the detector via import also collapses the two Python copies.
- Prior non-blocking #4 (log label) and #6 (FP boundary test) are also resolved.

### Cross-repo blast radius — verified clean
Each claim independently traced and adversarially re-checked:
- **divine-mobile:** renders first-party VTT verbatim (`SubtitleService.parseVtt`, whitespace-trim only, no sanitizer); an empty `WEBVTT` body yields zero cues → `subtitle_overlay.dart` returns `SizedBox.shrink()` (no crash). `TranscriptSanitizer` is not on `main` (only the closed/unmerged `fix-issue-3737`), so the backend stays the single home for this fix.
- **divine-web:** `useSubtitles.ts` passes the body only to `parseVtt`; empty VTT → 0 cues → overlay never mounts; no raw re-exposure (`vttParser.test.ts` 18/18).
- **divine-moderation-service:** `pipeline.mjs` gates `classifyText`/`extractTopics` on `plainText.trim().length > 0`; `parseVttText("WEBVTT\n\n")` → `""`, so an emptied VTT feeds nothing downstream (a net improvement); the cron reprocess path is also guarded.
- **Serving path is terminal:** the emptied VTT is uploaded, reported `status=complete`, and edge-cached (`public, max-age=31536000, immutable`; the completion webhook purges + re-warms with the empty body). No automatic re-transcription — `should_eagerly_trigger_transcription` fires only for `None`/`Pending`, create-job short-circuits on `Complete` unless `force`, and the repair scripts are manual CLIs (no scheduler). Recovery requires a manual `force=true` repair.
- **#134 attribution:** the leaked phrases (`single JSON array`, `extra text outside of the JSON`, `follow the schema provided in the context`) appear in no divine prompt or source — only in this PR's detector/tests and in mobile's investigation notes. `build_gemini_prompt` never contained them (git `-L` history; the prompt asks for a JSON object, the leak says JSON array). They are Vertex `responseSchema` boilerplate. (Minor: 2 historical prompt versions, not 3 — immaterial.)

### Non-blocking
1. **Residual over-fire on generic markers (worth a follow-up).** The overlapping-marker mechanism is fixed, but `follow the schema` + `provided in the context` (and `output requirements` + `follow the schema`) still fire on ordinary technical speech — e.g. *"Our API output requirements changed, so please follow the schema in the new docs."* drops permanently. Since the drop is terminal, each false positive is permanent caption loss until a manual repair. This is the inherent substring-matching tradeoff the team already accepted; flagging because the failure direction is non-recoverable and the fix is small (details inline). Reasonable to land as-is and tighten in a follow-up.
2. **Rust/Python parity gap on C0 control chars** (inline) — low practical impact, but breaks the byte-for-byte parity the three copies rely on.
3. **The new tests don't run in CI** (inline) — `cloud-run-transcoder` is excluded from the Cargo workspace and `ci.yml` has no transcoder/Python step, so the regression protection and parity this PR adds aren't enforced. A CI step + a marker-parity assertion would make prior non-blocking #5 durable.
4. **`--check-empty` re-trigger churn** — author deferred this; agreed it's a reasonable separate follow-up (operator-invoked, non-default).

Nice, well-targeted fix with strong test coverage and a clean cross-repo footprint.

NotThatKindOfDrLiz:
@dcadenas could you re-review when you have a chance? Your original blocking threads are now outdated and addressed, and I also pushed `633bcb2` to tighten the remaining generic-marker false positives, close Rust/Python C0 marker-scan parity, add marker-list parity coverage, and run the transcoder Rust/Python suites in CI.

Latest validation:
- `cargo test --manifest-path cloud-run-transcoder/Cargo.toml --locked` — 125 passed
- `python3 -m unittest discover -s tests -p 'test_*.py'` — 23 passed
- `cargo clippy --manifest-path cloud-run-transcoder/Cargo.toml --locked --all-targets` — exits 0 with pre-existing unrelated warnings only

dcadenas (approved):
🧹 caption cleanup with solid guardrails against false trims, plus repair scripts to backfill the bad VTTs.

PR #141, opened 2026-07-23
feat: server-side caption transcription for the editor (transcoder endpoint + edge forward)

hm21:
## Summary

Two services, one feature: a synchronous "audio -> WebVTT" path so the mobile editor gets server-quality captions before publish.

1. **`cloud-run-transcoder` - `POST /transcribe/audio`** (raw audio bytes -> WebVTT via the configured provider). Runs the hardened core `transcribe_audio_file` (silence guard, provider semaphore `TRANSCRIPTION_MAX_IN_FLIGHT`, `transcribe_with_fallback`, provider-specific parsing incl. `google_stt_v2`, and every caption-quality guard). Auth: a shared-secret header (see below); **fails closed (503)** when unset. Body cap 100 MB **plus** an ffmpeg `-t` decoded-duration cap.
2. **Fastly edge - `POST /transcribe`** rate-limits per pubkey and per IP, then forwards to the upload service's authenticated proxy, streaming the body and returning the upstream response as-is plus CORS.

Chain: edge `POST /transcribe` (rate limit) -> divinevideo/divine-upload-server#12 (Nostr-authed proxy, injects the shared secret) -> transcoder `POST /transcribe/audio`.

## Motivation

The mobile editor needs pre-publish captions without first uploading a finished video blob and waiting for the by-hash transcript pipeline. This gives the editor a synchronous server-side caption path while keeping the existing published-video transcription flow unchanged.

## Rate limiting (edge)

- Fixed-window limiter: per-pubkey **and** per-IP counters keyed by `floor(now/window)`, TTL of two windows, best-effort fail-open on KV outage.
- Over-limit returns `429` + `Retry-After` before any downstream ffmpeg/ASR work begins.
- Limits are tunable constants: **10 req/hour per pubkey**, **20 req/hour per IP**.
- Pure bucket/decision/`enforce` logic lives in `blossom-core`; Fastly KV is bound in the binary via a newtype.
- Residual, by design: the upload service is also directly reachable, so a caller that skips the edge is not throttled here. The transcoder's in-flight semaphore is the backstop for that path.

## Review feedback addressed

- Auth: `/transcribe/audio` requires `X-Divine-Transcribe-Secret`, injected by the upload-service proxy after it validates Nostr auth. It fails closed when unconfigured.
- Decoded-size bound: ffmpeg `-t MAX_TRANSCRIBE_AUDIO_SECONDS` caps PCM output so compressed inputs cannot expand unbounded.
- Provider fallback + provider-specific parsing + semaphore + quality guards: the raw-audio path runs through `transcribe_audio_file`.
- Edge headers: returns upstream response unchanged plus CORS, preserving `Content-Type` and `Retry-After`.
- Rate limiting: implemented at the edge with CAS-based KV writes.

**Deliberately deferred:** the by-hash `process_transcribe` still has its own inline copy of the core. Rerouting it through `transcribe_audio_file` is tracked in #147 because its GCS-lock/cooldown/webhook behavior needs staging validation before changing the production caption path. The raw-audio path is fully hardened.

## Review round 2 (addressed)

- Forged pubkeys could burn a victim's quota: IP is checked first, and pubkey is charged only after `verify_event_authenticity` confirms event id, signature, and non-expiry. Full action/kind validation remains downstream on the upload service.
- Concurrent requests bypassed the fixed window: replaced read-then-overwrite with compare-and-swap using `InsertMode::Add` and `if_generation_match`, retrying on `ItemPreconditionFailed`.
- Transcoder authenticated after buffering up to 100 MiB: the shared-secret check moved into route middleware (`require_transcribe_secret`) that runs before body extraction.
- 429 `Retry-After` not browser-readable: added it to the edge CORS `Access-Control-Expose-Headers`.

**Residual to validate on staging:** the Fastly-KV compare-and-swap binding (`Add` / `if_generation_match` -> `ItemPreconditionFailed`) cannot be exercised in `cargo test`; only the pure retry logic is. Confirm a concurrent burst is capped against deployed KV.

## Deploy prerequisites

1. **`TRANSCRIBE_SHARED_SECRET`** - provision a secret and set it on both this transcoder and divine-upload-server#12 (same value). Until then `/transcribe/audio` fails closed (503).

## Test plan

- `cargo test -p blossom-core`
- `cargo build --target wasm32-wasip1`
- Staging smoke: `POST media.divine.video/transcribe` with a signed kind-24242 `t=media` event and short WAV; confirm a burst returns `429` + `Retry-After`.

Follow-up: #147


NotThatKindOfDrLiz:
@hm21 I took over the cleanup items from review:

- Opened #147 to track folding the by-hash `process_transcribe` path onto `transcribe_audio_file` after staging validation.
- Added `TODO(#147)` at the duplicated-core note in `cloud-run-transcoder/src/main.rs`.
- Removed the unrelated formatting churn in the transcoder file.
- Restored/updated the PR description after linking the follow-up issue.

Validation passed locally:
- `cargo clippy --locked --all-targets --all-features` (passes; existing warnings remain)
- `cargo build --target wasm32-wasip1 --locked`

GitHub checks are green for the new head `e93c753`. Please review the takeover commit and merge when you are ready, after the deploy prerequisites/staging smoke in the PR body are satisfied.

NotThatKindOfDrLiz (approved):
Final takeover review: approved.

I re-reviewed the full diff after the cleanup commit. The prior correctness issues are addressed, the deferred transcription-core duplication is now tracked in #147 with an in-code TODO, and the unrelated formatting churn was removed. Remaining deploy/staging prerequisites are documented in the PR body.

src/main.rs
```diff
@@ -3673,6 +3678,103 @@ fn handle_upload_complete(mut req: Request, path: &str) -> Result<Response> {
...
+            return Some(too_many_requests(retry_after));
+        }
+    }
+
+    // Pubkey only after verifying the event's signature. `parse_auth_header`
+    // decodes attacker-controlled JSON, so charging an unverified pubkey would
+    // let anyone forge a victim's pubkey and burn their hourly budget. The
+    // upload service still does full action/kind validation downstream.
+    if let Ok(event) = crate::viewer_auth::parse_auth_header(auth_header) {
+        if crate::viewer_auth::verify_event_authenticity(&event, now).is_ok() {
```
dcadenas: Validate the transcription-specific authorization contract before charging the pubkey bucket here. `verify_event_authenticity` proves only that the event is authentic and not expired, so any authentic public event with the wrong kind or action can be replayed to consume the victim's hourly quota before the upload proxy rejects it. Require kind 24242, `t=media`, the required non-expired expiration, and the same server-scope rules as the upload proxy before incrementing `pk`, while keeping the IP limiter first. Add a regression test proving that an authentic wrong-kind or wrong-action event does not increment the pubkey counter, while a valid transcription authorization still receives `429` with `Retry-After` after the configured limit.

dcadenas (dismissed):
Bind pubkey quota charges to the transcription authorization contract before merge. Replayed authentic events must not be able to consume another user's hourly transcription quota before the upload proxy rejects the request.

hm21:
@dcadenas Fixed in `26d08ac`. The edge now gates the per-pubkey charge on `validate_transcribe_event`, which enforces the same contract as the upload proxy — kind 24242, `t=media`, a required and unexpired `expiration`, and a valid signature — before charging quota. A replayed authentic token for any other action now fails at the edge and charges nothing. Added 8 tests covering the replay and forge vectors.

src/viewer_auth.rs
```diff
@@ -62,6 +62,52 @@ pub fn parse_auth_header(auth_header: &str) -> Result<BlossomAuthEvent> {
...
+/// id + signature.
+///
+/// The edge uses this to bind a per-pubkey rate-limit charge to the pubkey that
+/// actually authorized transcription. Authenticity alone is not enough:
+/// `parse_auth_header` decodes attacker-controlled JSON, and a *replayed*
+/// authentic event signed for some other action (a captured `t=get` blob-view
+/// token, say) would pass a signature-only check and burn the victim's hourly
+/// transcription budget before the upload service rejects it. Requiring the
+/// transcription contract here means any event the upload service would reject
+/// also fails at the edge, charging no quota. Server-tag scoping and the rest
```
dcadenas: Make `validate_transcribe_event` enforce the complete upload-proxy contract before `enforce_transcribe_rate_limit` charges the `pk` bucket.
This validator currently accepts `t=MEDIA` through `eq_ignore_ascii_case` at `src/viewer_auth.rs:92-99` and explicitly defers `server`-tag scoping at `src/viewer_auth.rs:82-83`, while the companion proxy requires exact `t=media` and rejects disallowed server scopes.
As a result, a valid signed token can consume the signer’s 10/hour quota at `src/main.rs:3713-3718` even though the downstream request will always be rejected.
Require exact action matching and equivalent allowed-host parsing here, keep IP limiting first, and add a counter-level regression proving these rejected events do not write `pk` while a valid event does.

dcadenas (dismissed):
Before this can merge, make the edge and upload proxy enforce the same transcription authorization contract before charging the pubkey quota. Add the counter-level regression test so rejected uppercase-action and disallowed-server events cannot consume the signer’s hourly budget.

src/viewer_auth.rs
hm21: Fixed in ac5b989. Matched the edge gate to `handle_transcribe`'s contract: exact `t=media` and `server`-tag scoping to `[media.divine.video, upload.divine.video]`. Action + server-host parsing now lives in `blossom-core::transcribe` (host-run in CI, since the edge crate is only `cargo check`ed), and a counter-level regression asserts uppercase-action and foreign-server tokens never write the `pk` counter while a valid token is limited after its budget. IP limiter still runs first.

blossom-core/src/transcribe.rs
```diff
@@ -0,0 +1,107 @@
...
+    };
+    let authority_end = after_scheme
+        .find(['/', '?', '#'])
+        .unwrap_or(after_scheme.len());
+    let authority = &after_scheme[..authority_end];
+    let host_port = match authority.rfind('@') {
+        Some(index) => &authority[index + 1..],
+        None => authority,
+    };
+    let host = match host_port.find(':') {
```
dcadenas: Reject malformed ports in `server_tag_host` before the allow-list check. At this line, `https://media.divine.video:abc` is truncated to `media.divine.video`, so `src/viewer_auth.rs:108-116` accepts it and `src/main.rs:3712-3716` charges the pubkey while the upload proxy rejects the malformed authority. Match the proxy’s URL parsing semantics and add a counter-level regression proving this input leaves the `pk` counter unchanged while valid allowed hosts still pass.

dcadenas (changes requested):
The edge still accepts malformed `server` authorities that the upload proxy rejects. A signed request such as `https://media.divine.video:abc` can therefore consume a signer’s pubkey quota even though transcription cannot succeed. Align the edge parser with the proxy and add a regression proving malformed authorities do not write the `pk` counter before this PR merges.

blossom-core/src/transcribe.rs
hm21: Fixed in 734cdea. `server_tag_host` now parses with `url::Url` — the same parser `handle_transcribe` uses — so `https://media.divine.video:abc` fails to parse (non-numeric port), falls through to the whole value, and is rejected by the allow-list instead of being truncated to `media.divine.video`. `url` was already resolved in the edge dep tree. blossom-core covers the malformed-port host mapping (host-run in CI), and the edge counter-level regression now asserts a `:abc`-port token never writes the `pk` counter while a valid explicit port still passes.

dcadenas (approved):
The malformed-authority quota finding is resolved on the current head. Commit 734cdea parses the server tag with url::Url to match the upload proxy, rejects a non-numeric port before charging the pubkey counter, and adds both host-policy and counter-level regression coverage.

I also added the focused deployment prerequisite in c2be91b so deploy.sh preserves TRANSCRIBE_SHARED_SECRET through its complete --set-secrets replacement. I validated the script with bash -n, shellcheck, an exact single-binding assertion, and the current PR CI is green.

Alex, please review the takeover commit and merge when you are ready.

PR #142, opened 2026-07-23
feat(edge): forward POST /transcribe to the upload service

hm21:
## Summary

Adds a thin **`POST /transcribe`** route to the edge (`media.divine.video`) that forwards to the upload service's authenticated `/transcribe` proxy.

- Streams the request body (audio) straight through — never buffered, so it stays within Fastly Compute's ~5 MB WASM memory limit (same reason large uploads are proxied).
- Passes the `Authorization` header through unchanged; the upload service is the authority that validates the Blossom auth (`t=media`). The edge does not re-validate.
- Reuses the existing `upload_service` backend (already configured for `upload.divine.video`) — **no new Fastly backend needed**.
- Passes the upload service's response (WebVTT, or its error status) straight back. CORS preflight is covered by the existing global `OPTIONS` handler.

## Motivation

Completes the reachability chain for the mobile editor's pre-publish caption generation. The mobile Blossom client only ever addresses the `media.divine.video` edge (there is no `upload.divine.video` in the app), and the edge matches every route explicitly with no catch-all forward — so without this route the editor cannot reach the transcription endpoint at all.

Chain:
- **This PR** — edge `POST /transcribe` → forwards to ↓
- divinevideo/divine-upload-server#12 — authenticated `POST /transcribe` proxy → calls ↓
- divinevideo/divine-blossom#141 — transcoder `POST /transcribe/audio` (Gemini) → WebVTT

## Rate limiting

The rate-limiting requirement lives at the authenticated proxy (divine-upload-server#12), which knows the caller pubkey. This edge route is a pass-through and adds no throttle of its own.

## Test plan

- `cargo check` — clean (pre-existing warnings only).
- `cargo fmt --check` — my lines are clean (the repo has unrelated pre-existing fmt drift elsewhere, left untouched to keep this scoped).
- Not unit-tested: this is a thin I/O forward mirroring the existing `handle_upload_service_proxy`; the repo's unit tests cover pure path parsers, not proxy handlers. Reviewer should smoke-test `POST media.divine.video/transcribe` (signed kind-24242 `t=media` + short WAV) against staging once #12 is deployed and confirm WebVTT returns.

Closes #none

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -140,6 +140,9 @@ fn handle_request(req: Request) -> Result<Response> {
         (Method::PUT, "/upload") => handle_upload(req),
         // BUD-06: Upload requirements/pre-validation
         (Method::HEAD, "/upload") => handle_upload_requirements(req),
+        // Synchronous audio transcription — thin forward to the upload service's
+        // authenticated /transcribe proxy (which calls the transcoder).
+        (Method::POST, "/transcribe") => handle_transcribe_proxy(req),
         // Divine resumable control plane
         (Method::POST, "/upload/init") => handle_upload_init(req),
         (Method::POST, p) if p.starts_with("/upload/") && p.ends_with("/complete") => {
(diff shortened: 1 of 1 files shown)


hm21:
Consolidated into #141 — the edge forward and the transcoder endpoint are one feature in this repo, so they now ship as one PR with two commits.

Issue #144, opened 2026-07-24
Fix CI: repair Dockerfile.local cargo build + defer process-blob WIF auth


dcadenas:
## Fix

Two separate issues:

### 1. Dockerfile.local cargo build failure (unblocked)

**Root cause:** Root `Cargo.toml` declares `members = ["blossom-core"]` and `blossom-core = { path = "blossom-core" }`, but `Dockerfile.local` copies only `Cargo.toml`, `Cargo.lock`, and root `src/` before `cargo build --locked --target wasm32-wasip1 --release`. The path dependency must compile `blossom-core/src`, so Cargo fails: `failed to load manifest for workspace member .../blossom-core` → `failed to read .../blossom-core/Cargo.toml` → ENOENT.

**Fix:** Insert `COPY blossom-core/ blossom-core/` before the `RUN cargo build` step in `Dockerfile.local`'s builder stage. (This one-line fix already exists as commit bb499c9 on branch `feat/1949-fix-ci-repair-dockerfile-local-cargo-build-defer-p`.)

**Acceptance:**
- `docker` CI job passes — image builds for `linux/amd64,linux/arm64` and pushes to GHCR.
- `test`/Fastly jobs remain unchanged and green.

### 2. deploy-process-blob WIF auth failure (requires IaC change)

**Root cause:** The `deploy-process-blob` job uses `secrets.GCP_WORKLOAD_IDENTITY_PROVIDER`, `secrets.GCP_SERVICE_ACCOUNT`, `secrets.GCP_PROJECT_ID` and fails with an attribute-condition rejection. The iac-coreconfig WIF pools for POC/production/staging do **not** list `divinevideo/divine-blossom` in their `github_repositories_additional` — so the provider's attribute condition rejects the repo's OIDC token. This is not a deleted-pool issue; the repo was never authorized.

**Target:** process-blob belongs with the live media stack in `rich-compiler-479518-d2` (GCP project number 149672065768, classified as environment `poc` in iac-coreconfig). Use the POC WIF pool, not the iac "production" pool, despite this being the live/production Blossom media stack.

**Fix (two parts):**

[...]

PR #145, opened 2026-07-24
fix(ci): copy blossom-core into local image builder

dcadenas:
The local Docker image could not build after `blossom-core` became a Cargo workspace member.
This change gives the builder the missing crate so local and CI image builds can compile the WASM service again.

## Summary

- Copy the complete `blossom-core` crate into the Docker builder before the locked WASM build.

## Motivation

- The root Cargo manifest declares `blossom-core` as a workspace member and path dependency, but `Dockerfile.local` copied only the root manifests and `src/`. Cargo therefore stopped before dependency resolution because `/app/blossom-core/Cargo.toml` was missing.
- Copying the complete crate includes both its manifest and source while keeping unrelated services out of the build context.
- The separate process-blob authentication failure is not changed here because its target project and masked secret destinations are not confirmed.

## Related Issue

- Part of #144

## Validation

- [x] `cargo test --manifest-path cloud-run-upload/Cargo.toml --locked`

## Manual Test Plan / Notes

- Reproduced the original linux/amd64 missing-manifest failure before the change.
- Built both the linux/amd64 builder stage and complete runtime image after the change.
- Ran 99 `blossom-core` tests, 18 upload tests, 125 transcoder Rust tests, and 23 transcoder Python tests.
- Local Buildx advertises only amd64-family platforms, so linux/arm64 needs the approved GitHub Actions multi-architecture run.

## Visuals / API Examples

- [x] No visual change

## Public Artifact Review

### DIFF
--- a/Dockerfile.local
+++ b/Dockerfile.local
@@ -9,6 +9,7 @@ RUN rustup target add wasm32-wasip1
 WORKDIR /app
 COPY Cargo.toml Cargo.lock ./
 COPY src/ src/
+COPY blossom-core/ blossom-core/
 
 # Build the WASM binary
 RUN cargo build --locked --target wasm32-wasip1 --release


NotThatKindOfDrLiz (approved):
Reviewed the one-line Dockerfile.local change against the workspace/path dependency setup. Copying blossom-core into the builder stage is the right scoped fix for the locked WASM image build, and it does not affect the runtime image beyond the compiled wasm artifact.

PR #146, opened 2026-07-24
fix: durable derivative status delivery via Cloud Tasks + fix webhook 403

rabble:
## Summary

Make transcoder status callbacks (HLS + transcript) durable, and fix a pre-existing auth failure that had been silently dropping all of them.

**Delivery**
- Transcoder `send_status_webhook` / `send_transcript_status_webhook` now route through `deliver_status_payload`: enqueue to a Cloud Tasks queue when `STATUS_QUEUE_ENABLED`, else direct POST (rollback path). Cloud Tasks owns retries, backoff, and dead-lettering.
- Each callback carries a monotonic `generation` (send-time millis).

**Receiver ordering / idempotency**
- `BlobMetadata` gains `transcode_generation` / `transcript_generation`.
- Both status handlers drop callbacks whose generation is older than what's stored, returning `200 {"ignored":"stale_generation"}` so the queue stops retrying. Redelivery of the same update is a no-op.
- Backward compatible: a missing `generation` always applies (legacy senders, first update).

**Webhook auth fix**
- Transcoder status callbacks had been returning `403 Invalid webhook secret` for a long time — a secret mismatch between the transcoder's credential and the edge's stored value (the latter shared with, and canonical to, the moderation service). The system had been relying entirely on the `admin_sweep` GCS-reconciliation backstop.
- `validate_transcoder_webhook` now accepts a dedicated `transcoder_webhook_secret` in addition to the shared `webhook_secret`, letting the transcoder be re-keyed independently. The shared secret is untouched, so the moderation path is unaffected.

## Motivation

Status callbacks were not "occasionally lost" — they were being rejected on auth, so derivative status only ever advanced via reconciliation. This makes delivery durable and correct, and removes the silent auth failure.

## Linked issue

None yet — happy to file one if wanted. Design/rollout notes: `docs/superpowers/plans/2026-07-24-derivative-status-queue-simple.md`.

## Manual validation

- Auth: `POST /admin/transcode-status` with the new transcoder secret → `404` (past auth, unknown blob); `/admin/transcript-status` → `202`; wrong secret → `403`. No new `403`s from the redeployed transcoder revision.
- Build: transcoder unit tests + `blossom-core` tests pass; edge type-checks via `cargo check --tests` (Fastly SDK is WASM-only, so edge `#[test]`s can't link natively — per AGENTS.md).
- Rollout: queue created; enqueuer role granted; two Cloud Monitoring alerts (queue depth, failing delivery attempts); edge + transcoder deployed; `STATUS_QUEUE_ENABLED=true`.

## No visual change

Backend/infra only. No edge/upload/UI behavior change beyond the delivery transport and stale-drop.

### DIFF
--- a/blossom-core/src/delete_policy.rs
+++ b/blossom-core/src/delete_policy.rs
@@ -223,6 +223,8 @@ mod tests {
             transcript_retry_after: None,
             transcript_attempt_count: 0,
             transcript_terminal: false,
+            transcode_generation: None,
+            transcript_generation: None,
         }
     }
 
--- a/blossom-core/src/types.rs
+++ b/blossom-core/src/types.rs
@@ -76,6 +76,11 @@ pub struct BlobMetadata {
     /// Whether HLS generation has reached a terminal failure state
     #[serde(default)]
     pub transcode_terminal: bool,
+    /// Monotonic ordering token from the last applied transcode status callback.
+    /// A callback carrying a smaller generation is stale (superseded) and is
+    /// dropped by the receiver. Sourced from the transcoder's send-time clock.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub transcode_generation: Option<u64>,
     /// Video dimensions as "WIDTHxHEIGHT" (display dimensions after rotation)
     #[serde(skip_serializing_if = "Option::is_none")]
     pub dim: Option<String>,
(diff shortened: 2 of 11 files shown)


NotThatKindOfDrLiz:
Closing this in favor of #159, which carries the same durable derivative-status delivery work forward on a current base with a smaller scoped diff.

Thanks for doing the design work here. Two ideas from this PR are worth keeping explicit as the replacement work gets reviewed:

- Return a 200 response with an explicit stale-drop shape, e.g. {"ignored":"stale_generation"}, so Cloud Tasks treats intentionally ignored stale callbacks as terminal successes instead of retryable failures.
- Account for the clock-skew failure mode in any wall-clock generation scheme. If two Cloud Run instances disagree and an older failed callback gets a larger generation than a later successful callback, the blob can get stuck in the wrong terminal status.

The current branch is also now dirty against main, duplicates the dedicated transcoder webhook-secret auth fix already landed in #158, and has ordering/durability blockers around cached metadata reads, enqueue failure handling, and queue concurrency. Let’s review and finish the replacement branch instead.

PR #149, opened 2026-07-27
fix(edge): wire upload-service derivative failure fields into blob metadata

rabble:
## Summary

The Fastly edge now parses `transcode_error_code`/`transcode_terminal` and `transcript_error_code`/`transcript_terminal` from the upload service response and applies them in `publish_upload_service_upload`, replacing the hardcoded `None`/`false` placeholder (which carried a TODO for exactly this wiring).

## Motivation

Investigation into ~250 Sentry events/day from the transcoder (`FFmpeg failed` / `ffmpeg audio extraction failed`, issues 7371590104/7371590088):

- Affected blobs are not truncated uploads — e.g. one hot blob is a 5,016-byte synthetic test pattern (`ftyp` header + sequential bytes) re-uploaded by ~100 distinct npubs since Feb 2026.
- The upload service already detects this: probe fails → `classify_invalid_media_signal` → `invalid_media` terminal, skips its own transcode/transcript triggers, and includes the failure fields in its JSON response.
- But the edge discarded them (hardcoded `None`/`false`), so blob metadata stayed `Pending`. Every HLS playback then re-triggered transcoding (`trigger_on_demand_transcoding`) and every re-upload re-triggered transcription, producing endless doomed ffmpeg runs and Sentry noise.

With this change, invalid blobs are marked `Failed` + `terminal` at ingest, so playback/re-upload no longer schedules derivative work for them.

## Scope / out of scope

- New uploads via the proxied (non-resumable) path are covered.
- Resumable upload completion does not probe/classify yet — fields default to `None`/`false` with a TODO comment; follow-up.
- Legacy bad blobs already in GCS quiet down via the transcoder webhook path (invalid_media → terminal), which is already deployed; an optional backfill to mark known-bad hashes terminal can be done separately.

## Manual validation plan

- `cargo check --tests --locked` passes (matches CI edge gate); `cargo clippy --locked --all-targets --all-features` shows no new warnings.
- Added unit tests for `parse_upload_service_response` (failure fields present, absent/legacy response, missing sha256). Edge crate tests are compile-gated only in CI (fastly crate doesn't link on host), same as existing tests.
- After deploy: re-upload the known bad blob (5,016-byte synthetic MP4) and confirm blob metadata reports `transcode_status=failed`, `transcode_terminal=true`, and that no new `/transcode` call is triggered on playback; Sentry event rate for the two ffmpeg issues should drop to ~0.

No visual change.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -3288,6 +3288,47 @@ struct UploadServicePublishedUpload {
     content_type: String,
     thumbnail_url: Option<String>,
     dim: Option<String>,
+    transcode_error_code: Option<String>,
+    transcode_error_message: Option<String>,
+    transcode_terminal: bool,
+    transcript_error_code: Option<String>,
+    transcript_error_message: Option<String>,
+    transcript_terminal: bool,
+}
+
+/// Parse the upload service's JSON response into an UploadServicePublishedUpload.
+/// The upload service reports derivative validation failures (e.g. invalid_media
+/// for corrupt uploads / missing moov atom) so the edge can mark metadata
+/// terminal immediately instead of discovering it via transcode retries.
+fn parse_upload_service_response(
+    resp: &serde_json::Value,
+    content_type: &str,
+    fallback_size: u64,
+) -> Result<UploadServicePublishedUpload> {
+    Ok(UploadServicePublishedUpload {
+        sha256: resp["sha256"]
+            .as_str()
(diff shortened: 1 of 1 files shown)


NotThatKindOfDrLiz:
Closing in favor of #160, which keeps the upload-service derivative failure wiring but avoids trusting the current broad invalid_media terminal signal and preserves already-complete deduped metadata.

PR #153, opened 2026-07-27
fix(transcoder): preserve runtime configuration

dcadenas:
## Summary

Running the transcoder deployment script could remove live environment variables and secret bindings that the script does not know about.
This change makes deployment update the named settings instead of replacing the complete Cloud Run configuration.
It keeps the transcription-secret rollout from deleting unrelated production settings.

## What Changed

The deployment still writes every setting it owns, but it now preserves settings managed outside the script.
This follows up on the durable transcription-secret binding merged in #141.

- Changed `--set-env-vars` to `--update-env-vars`.
- Changed `--set-secrets` to `--update-secrets`.
- Kept the existing environment and secret mappings unchanged.

## Testing

The script parses successfully and contains exactly one additive environment update and one additive secret update.
No Cloud Run deployment was performed from this branch.

- `bash -n cloud-run-transcoder/deploy.sh`
- Exact assertions for one `--update-env-vars` and one `--update-secrets`
- Assertion that no `--set-env-vars` or `--set-secrets` remains
- `git diff --check`

`shellcheck` was not available locally.

## Risks

The change intentionally preserves runtime settings that are not represented in the script.
That is safer for the current manually managed Cloud Run service, but it also means removing an obsolete setting must be done explicitly.

No visual change.


dcadenas:
Admin merge rationale: this is a narrow production-safety fix. The current manual transcoder deployment command uses full-replacement flags, so its next run would remove live environment variables managed outside the script. This PR changes only those flags to additive updates and records the deployment guardrail in AGENTS.md. GitHub CI and the local CI-equivalent test suite, shell syntax check, exact-flag assertions, and diff validation are green. Merging does not deploy the transcoder; that remains a separate manual step.

PR #155, opened 2026-07-29
fix(media): serve public blobs whole so edge cache stores video bytes

### LINKED ISSUE #154: Video byte-range requests bypass edge cache entirely
## Problem

Video plays use HTTP range requests. For publicly cacheable media, the Compute app forwards the client's `Range` header to origin storage and returns a genuine `206 Partial Content`. The VCL edge cache cannot store these partial responses, so video bytes never enter the cache: every play on every device streams from origin.

Live probes (2026-07-29): repeating the identical range request on a video returns `x-cache: MISS` every time, including on objects confirmed cached via full GET. Range requests on small cached objects (thumbnails) correctly return `x-cache: HIT`, proving the edge can synthesize 206s from cached full objects.

Result: request hit rate looks healthy (~82%, dominated by thumbnails/small GETs) while byte-level origin offload is ~14% (videos dominate bytes). Origin egress and Compute streaming time are paid on nearly every play.

## Root cause

`handle_get_blob` / `handle_get_audio` forward the client `Range` to GCS (`download_blob_with_fallback`) and pass the 206 through. A cached 206 under the URL key would also be cache poisoning, so the correct fix is to fetch the full object and let the edge synthesize ranges.

## Scope

- Public media (Active status, non-admin): should fetch whole objects from origin so the edge caches them.
- Private/admin/age-gated traffic: must keep real 206s (never cached; forwarding ranges avoids whole-object amplification).
- HLS quality-variant paths (`/{hash}/720p` etc.): out of scope for the first fix — playlists use EXT-X-BYTERANGE against single-file TS, and strict HLS players receiving a full 200 to a mid-file range request need verification before changing behavior.

## Verification

After deploy, per the diagnosis report (docs/cache-offload-report-2026-07-29.md):

1. Repeat an identical range request on a public video: second request must return `x-cache: HIT`.
2. A different range on the same object must also return `x-cache: HIT`.
3. Byte-level origin offload should climb toward >80%; average miss time should collapse.

rabble:
## Summary

Fixes #154

Publicly edge-cacheable blob/audio GETs are now fetched **whole** from origin instead of forwarding the client's `Range` header, so the VCL edge cache stores the full object and synthesizes 206 Partial Content responses for subsequent anonymous range requests.

The range policy now uses the same cacheability facts as the response headers and VCL pass rule:

- anonymous, non-admin, Active media -> drop the origin range so the edge can cache the full object
- requests with `Authorization`, admin traffic, missing metadata, and non-Active media -> forward the range and mark the response private/no-store

## Motivation

Video players fetch via byte-range requests. Forwarding those ranges to origin produced genuine `206` responses that the edge cache cannot store — so video bytes never entered the cache and every play streamed from origin (request hit rate ~82% vs byte offload ~14%). Full diagnosis and probes: `docs/cache-offload-report-2026-07-29.md` (on `docs/cache-offload-report`).

## Behavior change (externally visible)

- **Cold anonymous public video/audio with a `Range` header**: now returns `200 OK` with the full object instead of `206` with the requested slice. This is RFC 7233-legal (servers MAY ignore Range), but should still get iOS/Android/web playback verification before deploy.
- **Requests carrying `Authorization`**: keep real origin 206s and are marked private/no-store, matching VCL's `return(pass)` behavior and avoiding whole-object re-download amplification on uncached passes.
- **Private/admin/age-gated/non-Active traffic**: keeps real 206s and is not publicly cached.
- **HLS quality-variant paths (`/{hash}/720p`, etc.)** are deliberately unchanged: playlists use EXT-X-BYTERANGE against single-file TS, and strict HLS player behavior with a full 200 needs verification first (tracked in #154).

No visual/client-UI change.

## Validation

- [x] `cargo check --tests --locked` — passed with existing dead-code warnings
- [x] `cargo clippy --locked --all-targets --all-features` — passed with existing warnings
- [x] `cargo test -p blossom-core --locked` — 115 passed
- [x] Unit coverage for origin range/cache policy: public anonymous drops range; Authorization/admin/Pending forward range; absent range stays absent
- [ ] **Post-deploy manual probes** (from #154): repeated identical range request on an anonymous public video must return `x-cache: HIT`; a different range on the same object must also HIT; an authorized public range request should continue to pass through without full-object amplification; watch byte-level origin offload and avg miss time for 48h.

Deploy plan after merge: `fastly compute publish` + purge, then the probes above. Rollback: reactivate previous Compute version.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -346,6 +346,22 @@ fn should_set_audio_content_length(status: StatusCode) -> bool {
     status != StatusCode::PARTIAL_CONTENT
 }
 
+/// Decide which Range header, if any, to forward to origin storage.
+/// Publicly cacheable objects are always fetched whole: an origin 206 response
+/// cannot be stored by the edge cache, so forwarding client ranges for public
+/// media would send every play to origin. Private/admin responses are never
+/// cached, so their ranges pass through and avoid re-downloading whole objects.
+fn origin_range_for_request(
+    client_range: Option<&str>,
+    publicly_cacheable: bool,
+) -> Option<&str> {
+    if publicly_cacheable {
+        None
+    } else {
+        client_range
+    }
+}
+
 fn clear_stale_audio_mapping(source_hash: &str, audio_hash: &str) {
     let _ = delete_audio_mapping(source_hash);
     match remove_from_audio_source_refs(audio_hash, source_hash) {
@@ -531,8 +547,18 @@ fn handle_get_blob(req: Request, path: &str) -> Result<Response> {
         .and_then(|h| h.to_str().ok())
         .map(|s| s.to_string());
 
+    // Public blobs are fetched whole so the edge cache stores the full object;
+    // only private/admin traffic gets partial responses from origin. Keep this
+    // in sync with the cache-header decision below.
+    let publicly_cacheable = !is_admin
+        && metadata
+            .as_ref()
+            .map(|meta| meta.status == BlobStatus::Active)
+            .unwrap_or(false);
+    let upstream_range = origin_range_for_request(range.as_deref(), publicly_cacheable);
+
     // Download from GCS with fallback to CDNs
-    let result = download_blob_with_fallback(&hash, range.as_deref())?;
+    let result = download_blob_with_fallback(&hash, upstream_range)?;
     let mut resp = result.response;
 
     // Surface provenance metadata if present on the origin object.
@@ -2167,6 +2193,7 @@ fn handle_get_audio(req: Request, path: &str) -> Result<Response> {
         .get_header(header::RANGE)
         .and_then(|h| h.to_str().ok())
         .map(|s| s.to_string());
+    let upstream_range = origin_range_for_request(range.as_deref(), !private_cache);
 
     // 5. Check cache: source->audio mapping
     if let Some(mapping) = get_audio_mapping(&hash)? {
@@ -2176,7 +2203,7 @@ fn handle_get_audio(req: Request, path: &str) -> Result<Response> {
                 BlossomError::Internal(format!("Failed to persist audio source refs: {}", e))
             })?;
             // Serve cached audio via redirect or proxy
-            let result = download_blob_with_fallback(&mapping.audio_sha256, range.as_deref())?;
+            let result = download_blob_with_fallback(&mapping.audio_sha256, upstream_range)?;
             let mut resp = result.response;
             add_audio_response_headers(
                 &mut resp,
@@ -2262,7 +2289,7 @@ fn handle_get_audio(req: Request, path: &str) -> Result<Response> {
     })?;
 
     // 8. Download and serve the audio
-    let result = download_blob_with_fallback(&audio_sha256, range.as_deref())?;
+    let result = download_blob_with_fallback(&audio_sha256, upstream_range)?;
     let mut resp = result.response;
     add_audio_response_headers(&mut resp, &hash, private_cache, &mime_type, size, duration);
     Ok(resp)
@@ -5831,7 +5858,7 @@ mod tests {
     use super::{
         backfill_batch_cursor, classify_audio_reuse_availability, decide_transcode_fetch_action,
         decide_transcript_fetch_action, error_response, is_alias_only_audio_blob,
-        is_quality_variant_path, parse_quality_variant_path,
+        is_quality_variant_path, origin_range_for_request, parse_quality_variant_path,
         parse_transcript_status_webhook_payload, should_delete_derived_audio_blob,
         should_eagerly_trigger_transcription, should_set_audio_content_length,
         upload_capability_headers, upload_control_host, upload_exposed_headers,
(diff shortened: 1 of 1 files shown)


dcadenas (approved):
🎥 whole blobs keep video bytes cacheable

src/main.rs
```diff
@@ -531,8 +547,18 @@ fn handle_get_blob(req: Request, path: &str) -> Result<Response> {
         .and_then(|h| h.to_str().ok())
         .map(|s| s.to_string());
 
+    // Public blobs are fetched whole so the edge cache stores the full object;
+    // only private/admin traffic gets partial responses from origin. Keep this
+    // in sync with the cache-header decision below.
+    let publicly_cacheable = !is_admin
```
NotThatKindOfDrLiz: This predicate misses the VCL cacheability gate. `vcl/recv.vcl` returns `pass` for any request with `Authorization`, but non-admin viewer auth on an Active public blob still makes this `true`, drops the client Range, and pulls/streams the whole object with no cache benefit. Please include `req.get_header(header::AUTHORIZATION).is_none()` or move this into a shared helper that models actual public edge-cacheability.

NotThatKindOfDrLiz: This also creates a second definition of public cacheability. The response cache headers below treat raw blobs as public only when metadata is Active, while audio uses `!requires_private_cache()` where Pending is public. Since the range decision and headers must agree, please centralize this as a metadata/request cacheability helper and use it for both decisions instead of carrying a manual keep-in-sync invariant.

src/main.rs
```diff
@@ -6182,6 +6209,25 @@ mod tests {
         ));
     }
 
+    #[test]
+    fn publicly_cacheable_media_drops_client_range_for_origin() {
+        assert_eq!(origin_range_for_request(Some("bytes=0-1023"), true), None);
```
NotThatKindOfDrLiz: These tests only assert the helper's boolean passthrough behavior, so they would not catch the actual regressions here: Authorization-bearing public requests and status/predicate divergence between blob and audio. Please test the cacheability decision inputs directly, or change the helper signature to take the real inputs and cover Active/Pending/private/admin/Authorization cases.

NotThatKindOfDrLiz (changes requested):
Requesting changes. The cacheability policy needs to match the VCL layer before this ships: requests carrying Authorization are pass-through in VCL, so they should not drop Range and whole-fetch from origin. The new tests also need to cover that predicate, not only the identity helper.

NotThatKindOfDrLiz (approved):
Approved after takeover fixes. The current head aligns range dropping with VCL cacheability, keeps Authorization/admin/non-Active traffic ranged/private, adds predicate-level coverage, and CI is green. Do not skip the post-deploy media probes in the PR body.

NotThatKindOfDrLiz:
@rabble I took over the PR and pushed the fixes requested in review.

What changed:
- rebased the branch onto current main
- made blob/audio range dropping depend on actual edge cacheability, including VCL's Authorization pass rule
- unified blob/audio private-cache decisions for Active vs Pending/non-Active media
- kept authorized/admin/private/non-Active traffic forwarding ranges to avoid whole-object amplification on pass-through requests
- stopped audio responses from overwriting full-response Content-Length when a stored mapping has size 0, and fail new extraction results that return no size instead of persisting 0
- replaced the tautological helper tests with cases for anonymous Active, Authorization, admin, Pending, and absent Range

Validation:
- cargo clippy --locked --all-targets --all-features (passes; existing warnings remain)
- cargo test -p blossom-core --locked (115 passed)
- GitHub checks are green on 7869640

Please review the takeover changes and merge when you are ready. Post-deploy, the PR body's curl/cache probes should still be run before calling the rollout good.

rabble:
Closing this as superseded by the cache-fill behavior now on main and the current authenticated-media cache-boundary work. The edge already removes Range for outer cache fills and stores the complete object; this branch would additionally classify authenticated media reads as private/pass-through, which conflicts with the current direction of keeping permitted media responses safely cacheable. The useful full-object intent is retained without merging this older policy.

rabble (commented):
Second-model review, with the context of last night's cache work (this comment comes from the author's gh identity, so comment rather than request-changes — but please treat the first finding as blocking).

**1. Blocking: the PR is three weeks stale and now conflicts with shipped behavior.** `blob_response_requires_private_cache` treats `Pending` as private (`status != Active`, with a test enshrining it). Since last night (Compute v339), `Pending` is deliberately *publicly* cacheable — that change is what collapsed the pass share from ~48% to ~6% this hour, and it measured ~72% of sampled popular videos as `Pending`. Merging this as-is would flip the bare-blob and audio routes for `Pending` back to `private, no-store` and re-create the byte-offload hole. The remediation is mechanical: build the policy on `BlobStatus::requires_private_cache()` (Restricted | AgeRestricted) instead of `!= Active`, matching the rest of current `main`. The diff also still calls `download_blob_with_fallback` — current `main` routes this through `download_blob_read_through` (FOS mirror + #209's mirror-response check), so it needs a real rebase, not just a tweak.

**2. Worth stating explicitly before merge: what this adds now.** The motivating gap from the 07-29 report (origin 206s can't be cached) was closed the same day by the `vcl_miss` Range strip, which is still live. The marginal value of doing it at the Compute layer is defense-in-depth: it keeps whole-object fills correct if the VCL snippet is ever lost in a VCL edit (the drift between the two deploy paths bit us repeatedly this week), and it covers direct-Compute paths. That rationale is good enough to land it — but the PR body should say so, since as written it reads as the primary fix for a problem already fixed elsewhere.

**3. The flagged client risk has mostly already shipped.** "Cold anonymous range request gets a 200 full object" has been the effective production behavior on outer-cache misses since the strip went live on 07-29 (with do_stream since 08-11). So iOS/Android/web have been riding this for weeks on the miss path; this PR extends it to Compute-direct paths. Playback verification is still worth doing, but the risk is narrower than the body implies.

**4. Upside worth noting:** with #209 live, whole-object fetches (`had_range == false`) are write-back eligible, so this policy also heals the FOS mirror for public content, and the mirror size check composes cleanly with both branches of `origin_range_for_request`.

Also: the `src/main.rs` unit tests here predate #175 — after a rebase, CI runs the edge-crate tests under Viceroy, so these would actually execute now. Please rebase onto current main, switch to `requires_private_cache()`, re-run the suite, and update the body with the defense-in-depth framing; happy to re-read after that.

PR #158, opened 2026-07-29
fix(edge): accept dedicated transcoder webhook secret

### LINKED ISSUE #156: Fix transcoder status webhook authentication
## Problem
Transcoder status callbacks for HLS and transcript updates need to authenticate independently from the moderation webhook secret. Current main only accepts the shared webhook_secret for /admin/transcode-status and /admin/transcript-status, which couples transcoder rotation to moderation and caused callback 403s when the credentials diverged.

## Impact
Derivative status can fail to advance through webhook delivery, leaving status reconciliation to backstops instead of the primary callback path.

## Acceptance criteria
- /admin/transcode-status accepts webhook_secret and transcoder_webhook_secret.
- /admin/transcript-status accepts webhook_secret and transcoder_webhook_secret.
- Existing admin bearer validation behavior is preserved.
- Header parsing remains covered by tests.

## Technical notes
Keep moderation on webhook_secret. Do not rotate or rename existing secrets in code.

NotThatKindOfDrLiz:
## Summary
- share bearer token parsing/Secret Store validation for admin-style bearer checks
- let transcoder status callbacks authenticate with either webhook_secret or transcoder_webhook_secret
- keep moderation on the existing shared webhook_secret path

## Motivation
Closes #156. The transcoder status endpoints currently accept only the shared moderation webhook secret, which couples transcoder credential rotation to moderation and can make status callbacks 403 when the credentials diverge.

## Validation

## No visual change
Backend auth path only.

### DIFF
--- a/src/admin.rs
+++ b/src/admin.rs
@@ -56,39 +56,37 @@ pub fn validate_admin_auth(req: &Request) -> Result<()> {
 /// Accepts either admin_token or webhook_secret from the Fastly Secret Store.
 /// Both are checked independently — webhook_secret is used by divine-moderation-service.
 pub fn validate_bearer_token(req: &Request) -> Result<()> {
-    let provided = req
-        .get_header(header::AUTHORIZATION)
+    let provided = extract_bearer_token(req)
+        .ok_or_else(|| BlossomError::AuthRequired("Admin authentication required".into()))?;
+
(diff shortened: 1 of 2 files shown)


dcadenas (approved):
🔒 dedicated webhook secret support is ready to ship

PR #159, opened 2026-07-29
fix(transcoder): deliver derivative status through durable queue

### LINKED ISSUE #157: Deliver derivative status callbacks through a durable queue
## Problem
HLS and transcript status callbacks are currently direct POSTs from the transcoder to the edge. A transient request failure can lose the status update, and retrying callbacks needs an ordering guard so older status cannot overwrite newer status.

## Impact
Derivative status can become stale or rely on reconciliation instead of durable primary delivery.

## Acceptance criteria
- Transcoder can enqueue status callbacks to Cloud Tasks behind a disabled-by-default STATUS_QUEUE_ENABLED flag.
- Direct POST remains the rollback path when the flag is false.
- Each callback carries a generation token and the edge ignores stale generations.
- Generation checks do not read through the POP-local metadata cache.
- Queue dispatch is versioned with max-concurrent-dispatches=1.
- CreateTask failures are retried and emit an alertable log signal.
- Queue setup and Cloud Run env names are versioned in the repo.

## Technical notes
This should not include unrelated research or superseded design docs. Queue enablement in production should happen only after the setup script has created/updated the queue and IAM grant.

NotThatKindOfDrLiz:
## Summary
- add disabled-by-default Cloud Tasks delivery for transcode and transcript status callbacks
- stamp callbacks with monotonic millisecond attempt/event generation tokens and ignore stale or duplicate generations at the uncached metadata update boundary
- have edge-originated transcode and transcript status writes participate in the same generation sequence so self-heal writes cannot be regressed by delayed queued callbacks
- reject malformed webhook generations and ignore missing generations once versioned state exists, with a trimmed/case-insensitive Fastly config-store rollback switch for legacy generation-less transcoder images
- retry CreateTask with deterministic task names, a shared bounded HTTP client, 409-as-success idempotency, and direct POST fallback after exhausted enqueue attempts
- version queue env vars in deploy.sh plus a setup script that derives the same project, location, queue name, service account, serialized queue config, resume behavior, and queue-scoped IAM grant
- document the rollout, alerting contract, serialized throughput ceiling, fallback ordering limitation, receiver rollback switch, max-attempt exhaustion monitoring, missing-blob reconciliation response, and bearer-secret storage tradeoff

## Motivation
Direct POST status callbacks can be lost on transient failures. Queue delivery needs a receiver-side ordering guard and serialized dispatch so older status cannot overwrite newer status.

Closes #157.

## API behavior
Missing derivative metadata callbacks are now acknowledged instead of retried forever on the serialized queue. Transcode and transcript status callbacks for a blob whose metadata no longer exists return HTTP 202 with a reconciliation body, for example:

```json
{
  "success": true,
  "sha256": "<sha256>",
  "reconciliation": "pending",
  "message": "Transcode status accepted for later reconciliation",
  "status": "failed"
}
```

## Validation
- cargo test --manifest-path cloud-run-transcoder/Cargo.toml --locked
- cargo clippy --locked --all-targets --all-features (passes with existing warnings)
- cargo clippy --manifest-path cloud-run-transcoder/Cargo.toml --locked --all-targets --all-features (passes with existing warnings)
- bash -n cloud-run-transcoder/deploy.sh
- bash -n cloud-run-transcoder/setup-status-queue.sh

Note: `cargo test --tests --locked` compiles locally but does not link against the Fastly host shims on macOS; `cargo check --tests --locked` compiles those tests.

## No visual change
Backend delivery, edge metadata ordering, and operational config only.


cloud-run-transcoder/src/main.rs
```diff
@@ -5625,6 +5698,166 @@ fn build_transcode_status_webhook_payload(
...
+
+async fn deliver_status_payload(
+    config: &Config,
+    hash: &str,
+    label: &str,
+    status: &str,
+    target_url: &str,
+    mut payload: serde_json::Value,
+) {
+    let generation = current_generation_ms();
```
dcadenas: Assign the generation when the derivative attempt starts, not when this callback is sent. This timestamp conflates attempt order with callback order and retry identity, so an older HLS attempt that finishes later can look newer. Carry a stable attempt ID plus an explicit event sequence through processing and terminal callbacks, reject or quarantine missing or malformed generations once versioned state exists, and make equal-generation terminal retries idempotent. Add tests for an older attempt finishing after a newer one, normal processing-to-terminal progress, missing or malformed generations, and duplicate failed callbacks.

dcadenas (changes requested):
Give each derivative attempt a stable identity before work starts and carry it through every status callback. The current send-time timestamp can let an older overlapping attempt overwrite newer metadata, while missing or equal generations can replay terminal failure side effects. Define and test the ordering, legacy or malformed-generation rollout, and idempotent duplicate handling before merge.

NotThatKindOfDrLiz:
Takeover update:

- rebased the PR onto `main` and retargeted it so CI runs against the real merge base
- changed derivative generations from send-time timestamps to monotonic attempt tokens plus per-event sequencing, so older overlapping attempts cannot outrank newer work
- made duplicate failed callbacks with the same generation idempotent, so Cloud Tasks redelivery cannot replay attempt-count increments into terminal state
- added direct POST fallback after exhausted `CreateTask` retries
- added focused tests for generation ordering, duplicate/stale generation decisions, and duplicate ignore responses
- expanded `docs/derivative-status-queue.md` with the fallback, throughput ceiling, and bearer-secret storage notes

Validation:

- `python3 -m unittest discover -s tests -p 'test_*.py'` from `cloud-run-transcoder`
- `cargo clippy --locked --all-targets --all-features` (passes with existing warnings)
- `bash -n cloud-run-transcoder/setup-status-queue.sh`
- GitHub CI `test` passed on `b7b02be`

Note: I also tried `cargo test --locked metadata::tests`; local linking fails against Fastly host symbols on macOS, while `cargo check --tests --locked` compiles those tests.

NotThatKindOfDrLiz:
Takeover update on `47bdab8`:

- rebased the PR onto current `main`, resolving the `src/main.rs` test import conflict
- added the receiver-side rollback switch `REQUIRE_DERIVATIVE_STATUS_GENERATION=false` for rolling back to older transcoder images that do not send `generation`
- collapsed the duplicated ignored-generation response builders and transcode/transcript match handling into shared helpers
- changed missing transcode metadata callbacks to 202 reconciliation responses so Cloud Tasks does not retry deleted/missing blobs on the serialized queue
- aligned queue setup/deploy/runtime defaults around `GCP_PROJECT_ID`, `STATUS_QUEUE_LOCATION`, and `STATUS_QUEUE_NAME`
- replaced per-attempt status HTTP clients with one shared bounded client that keeps the 10s timeout
- documented rollback, queue config alignment, and max-attempt exhaustion monitoring

Validation:

- `cargo clippy --locked --all-targets --all-features` (passes with existing warnings)
- `cargo clippy --manifest-path cloud-run-transcoder/Cargo.toml --locked --all-targets --all-features` (passes with existing warnings)
- `bash -n cloud-run-transcoder/setup-status-queue.sh`
- GitHub checks on `47bdab8`: `test`, `semantic-pull-request`, and CLA passed; deploy jobs skipped as expected

@dcadenas please re-review when you have a chance. I did not approve or merge.

NotThatKindOfDrLiz:
Takeover update on `2bf4076`:

- rebased the branch onto current `main` (`a47d184`)
- fixed edge-originated transcode status writes so self-heal/repair updates now participate in derivative generation ordering
- made equal-generation callbacks idempotent for all statuses, not only failed attempts
- made `deploy.sh` and `setup-status-queue.sh` derive `PROJECT_NUMBER` only when `SERVICE_ACCOUNT` is not explicitly set
- added Cloud Tasks request-path tests for `409 Conflict` idempotency and CreateTask retry behavior
- changed the rollback/config flag parsing to trim and case-fold values, sharing it with `ENABLE_PHYSICAL_DELETE`
- documented the missing-blob HTTP 202 reconciliation response in the PR body and rollout doc
- removed the unrelated formatting-only churn called out in review

Validation:

- `cargo clippy --locked --all-targets --all-features` (passes with existing warnings)
- `cargo clippy --manifest-path cloud-run-transcoder/Cargo.toml --locked --all-targets --all-features` (passes with existing warnings)
- `bash -n cloud-run-transcoder/deploy.sh`
- `bash -n cloud-run-transcoder/setup-status-queue.sh`

I did not approve or merge. @dcadenas please re-review when you have a chance.

NotThatKindOfDrLiz:
Takeover update on `e973624`:

- stamped edge-originated transcript status writes with derivative generations, matching the existing transcode self-heal behavior so delayed queued transcript callbacks cannot regress repaired state
- removed the dead `generation_rejection` parameter and the zero-caller `update_transcode_status_with_size` helper
- aligned `deploy.sh` service-account derivation with the same `GCP_PROJECT_ID` used by queue setup/runtime config
- created #168 and added `TODO(#168)` for replacing queued bearer-header auth with Cloud Tasks OIDC
- updated the PR body to include transcript self-heal writes in the generation-ordering summary

Validation:

- `cargo clippy --locked --all-targets --all-features` (passes with existing warnings)
- `cargo clippy --manifest-path cloud-run-transcoder/Cargo.toml --locked --all-targets --all-features` (passes with existing warnings)
- `bash -n cloud-run-transcoder/deploy.sh`
- `bash -n cloud-run-transcoder/setup-status-queue.sh`
- `git diff --check`
- GitHub checks on `e973624`: `test`, `semantic-pull-request`, and CLA passed; deploy jobs skipped as expected

I did not approve or merge. @dcadenas please re-review when you have a chance.

src/metadata.rs
```diff
@@ -263,13 +263,18 @@ pub fn update_blob_status(hash: &str, status: BlobStatus) -> Result<()> {
 
 /// Update transcode status for a video blob
 pub fn update_transcode_status(hash: &str, status: crate::blossom::TranscodeStatus) -> Result<()> {
+    let generation = Some(edge_transcode_status_generation(status));
```
dcadenas, line 266: **Edge self-heal writes will outrank a transcoder callback from an older attempt.**

The transcoder stamps its generation when the attempt starts; this one is stamped at write time. Job start is always earlier, so when the self-heal at `src/main.rs:743` fires in the window between the playlist landing in GCS and the queued callback being applied, that callback is rejected as stale and the `dim` it carried is dropped.

Low impact today. `dim` is normally set from the upload service at publish time (`src/main.rs:3543`) — all 66 videos among the 200 most recent blobs already had it, including ones that had not started transcoding. So this needs a video that missed that path plus a view inside the window. Non-blocking, but the two sides are measuring different events, and that will surface again.

src/metadata.rs
```diff
@@ -1249,3 +1353,110 @@ pub fn delete_audio_source_refs(audio_hash: &str) -> Result<()> {
         ))),
     }
 }
+
+#[cfg(test)]
+mod tests {
```
dcadenas, line 1358: **These tests do not run in CI.**

They do run locally under Viceroy. `CARGO_TARGET_WASM32_WASIP1_RUNNER="$HOME/.config/fastly/viceroy run -C fastly.toml --" cargo test --target wasm32-wasip1 --bin fastly-blossom` passes all 13 generation tests, and 122 across the crate. Plain `cargo test` fails to link only because it defaults to the host target — so the macOS note in the description reads as a platform limit when it is really just the default target.

CI runs `cargo check --tests --locked` for this crate, and `cargo test` only for `blossom-core`, `cloud-run-upload`, and `cloud-run-transcoder`, so none of these assertions execute there. Adding the Viceroy invocation as a CI step would cover the whole crate, not just this file.

docs/derivative-status-queue.md
```diff
@@ -0,0 +1,78 @@
...
+
+Run `cloud-run-transcoder/setup-status-queue.sh` before enabling the flag. The
+script uses the same `GCP_PROJECT_ID`, `STATUS_QUEUE_LOCATION`, and
+`STATUS_QUEUE_NAME` values as `cloud-run-transcoder/deploy.sh`, and creates or
+updates the queue with `--max-concurrent-dispatches=1`.
+Serialized dispatch is part of the ordering contract: the edge receiver stores
+`transcode_generation` and `transcript_generation` on blob metadata and ignores
+callbacks whose generation is older than the stored value.
+
+Generation values use a monotonic derivative-attempt token plus a small
```
dcadenas, line 18: "monotonic" holds within a single transcoder process — `LAST_STATUS_GENERATION` is a process-local static. Across Cloud Run instances the ordering comes from the wall clocks agreeing. Worth saying so here, since a reader would reasonably assume more.

src/main.rs
```diff
@@ -3235,6 +3258,7 @@ fn handle_upload(mut req: Request) -> Result<Response> {
         transcode_retry_after: None,
         transcode_attempt_count: 0,
         transcode_terminal: false,
+        transcode_generation: None,
```
dcadenas, line 3261: A re-upload clears the stored generation, so a callback still in flight from an earlier attempt has nothing to compare against and will apply. Narrow, given content-addressed hashes and the short window, but worth a note if you want the guard airtight.

dcadenas (approved):
✅ generations now come from the attempt, not the callback — four non-blocking suggestions inline.

Issue #151, opened 2026-07-27
Resumable uploads bypass media probe and derivative classification


rabble:
## Problem

The direct upload path in `cloud-run-upload/src/main.rs` probes video bytes and classifies invalid media (`classify_invalid_media_signal` → `invalid_media` terminal, skips transcode/transcript triggers, reports fields in the upload response). The resumable upload path (`cloud-run-upload/src/resumable.rs`) does **none** of this: `CompleteUploadResponse` returns no `thumbnail_url`/`dim`, no probe runs, and no derivative failure classification happens.

Result: invalid video (corrupt container, missing moov atom, synthetic junk) uploaded via resumable sessions enters GCS unvalidated, then fails repeatedly in the transcoder (Sentry noise, wasted compute) until the transcoder webhook marks it terminal — one full failure cycle per blob, plus no thumbnail/dim for legitimate resumable uploads.

The edge marks this explicitly with a TODO in `src/main.rs` (resumable completion call site: "resumable completion does not probe/classify invalid media yet").

## Proposal

- In `complete_session`, after finalizing the object, run the same sanitize/probe/classify pipeline used by the direct path (bytes available via the finalized temp object).
- Extend `CompleteUploadResponse` (and `ResumableUploadCompleteResponse` in `blossom-core/src/types.rs`) with `thumbnail_url`, `dim`, and the derivative failure fields (`transcode_error_code/message/terminal`, `transcript_error_code/message/terminal`) so the edge can mark metadata terminal immediately, matching the direct path wired in #149.
- Skip transcode/transcript triggers when classification fires.

## Context

Found during investigation of ~250 Sentry events/day from the transcoder on invalid source media. Related: #149.

PR #160, opened 2026-07-29
fix(edge): safely record upload derivative failures

NotThatKindOfDrLiz:
## Summary

- Parse derivative failure fields returned by the upload service on proxied uploads.
- Record upload-service derivative errors without trusting broad `invalid_media` terminal flags.
- Preserve already-complete HLS/VTT metadata during deduped reuploads.
- Reset prior failed derivative state on clean reupload, and track the resumable gap with `TODO(#151)`.

## Motivation

The upload service already returns derivative validation fields, but the edge was still hardcoding them to `None`/`false`. That left known-bad uploads in a retry loop. PR #149 wired the fields through, but trusted `terminal: true` too broadly and could permanently block recoverable browser-recorded WebM uploads or downgrade already-complete blobs.

This keeps the useful part of that fix while avoiding the permanent-terminal regression.

## Linked Issue

Follow-up: #151 tracks resumable uploads not yet returning derivative failure fields.

## Validation

- `git diff --check`

`cargo test --locked` was attempted, but native linking fails on Fastly host symbols such as `_header_insert` and `_pending_req_wait_v2`; the test targets compile under `cargo check --tests --locked`.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -3293,6 +3293,66 @@ struct UploadServicePublishedUpload {
     content_type: String,
     thumbnail_url: Option<String>,
     dim: Option<String>,
+    transcode_error_code: Option<String>,
+    transcode_error_message: Option<String>,
+    transcode_terminal: bool,
+    transcript_error_code: Option<String>,
+    transcript_error_message: Option<String>,
+    transcript_terminal: bool,
+}
+
+fn trusted_upload_service_terminal_derivative_error(
+    error_code: Option<&str>,
+    upload_service_terminal: bool,
+) -> bool {
+    upload_service_terminal
+        && matches!(
+            error_code,
+            // Do not include invalid_media here. The upload service currently
+            // uses that broad code for sanitize/probe failures that can be
+            // recoverable by the transcoder's full re-encode path.
+            Some("unsupported_media_type")
+        )
+}
+
+fn parse_upload_service_response(
+    resp: &serde_json::Value,
+    content_type: &str,
+    fallback_size: u64,
+) -> Result<UploadServicePublishedUpload> {
+    let transcode_error_code = resp["transcode_error_code"].as_str().map(|v| v.to_string());
+    let transcript_error_code = resp["transcript_error_code"]
+        .as_str()
+        .map(|v| v.to_string());
+    let transcode_terminal = trusted_upload_service_terminal_derivative_error(
+        transcode_error_code.as_deref(),
+        resp["transcode_terminal"].as_bool().unwrap_or(false),
+    );
+    let transcript_terminal = trusted_upload_service_terminal_derivative_error(
+        transcript_error_code.as_deref(),
+        resp["transcript_terminal"].as_bool().unwrap_or(false),
+    );
+
+    Ok(UploadServicePublishedUpload {
+        sha256: resp["sha256"]
+            .as_str()
+            .ok_or_else(|| BlossomError::Internal("Missing sha256 in Cloud Run response".into()))?
+            .to_string(),
+        size: resp["size"].as_u64().unwrap_or(fallback_size),
+        content_type: content_type.to_string(),
+        thumbnail_url: resp["thumbnail_url"].as_str().map(|v| v.to_string()),
+        dim: resp["dim"].as_str().map(|v| v.to_string()),
+        transcode_error_code,
+        transcode_error_message: resp["transcode_error_message"]
+            .as_str()
+            .map(|v| v.to_string()),
+        transcode_terminal,
+        transcript_error_code,
+        transcript_error_message: resp["transcript_error_message"]
+            .as_str()
+            .map(|v| v.to_string()),
+        transcript_terminal,
+    })
 }
 
 fn extract_authorization_header(req: &Request) -> Result<String> {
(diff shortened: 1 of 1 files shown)


rabble (commented):
Reviewed. The core direction is right — this fixes the real bug (upload service has been returning these fields since `classify_invalid_media_signal` landed, and the edge was throwing them away) without reintroducing #149's permanent-terminal regression. The decision helpers are well-factored and testable.

I pushed a follow-up commit (a71ddde) with two fixes rather than blocking on them. Please sanity-check it.

### Fixed in a71ddde

**1. Reupload silently downgraded an existing terminal verdict.**

`should_record_upload_service_transcode_failure` returns `true` when the current status is `Failed`, including a `Failed` that the transcoder webhook already marked `transcode_terminal: true`. The body then did:

```rust
metadata.transcode_terminal = transcode_terminal;
```

Since `trusted_upload_service_terminal_derivative_error` clamps `invalid_media` to `false`, any reupload of a permanently-rejected blob reset `transcode_terminal` to `false` and cleared `transcode_retry_after`, dropping it back into the retry loop that `decide_transcode_fetch_action` was supposed to have terminated. Now `metadata.transcode_terminal || transcode_terminal`. Same for transcript.

**2. Dedupe upload stomped an in-flight `Processing` derivative.**

The guard only excluded `Complete`, so a dedupe upload arriving while the transcoder was mid-run flipped the status to `Failed` and set `transcode_retry_after = None`, causing a second trigger on the next HLS fetch while the first was still running. The in-flight run's webhook is authoritative — `Processing` is now excluded alongside `Complete`.

Verified with `cargo check --tests --locked` and `cargo clippy --locked --all-targets --all-features`, both clean.

### Non-blocking, worth a note

**3. The terminal allowlist is currently dead code.**

`trusted_upload_service_terminal_derivative_error` only trusts `unsupported_media_type`, but `cloud-run-upload/src/main.rs:1761` (`classify_invalid_media_signal`) only ever emits `invalid_media` — both branches hardcode that string. So `transcode_terminal`/`transcript_terminal` parse to `false` 100% of the time today. That's the safe default and I'm not asking to change it, but the PR description reads as though terminal is preserved for some cases when in practice it never is. Worth either a comment saying the allowlist is forward-looking, or an issue to make the upload service emit a narrower code for genuinely unrecoverable containers.

**4. These tests never run in CI.**

`.github/workflows/ci.yml` runs `cargo check --tests --locked` for the edge crate and `cargo test` only for `blossom-core`, `cloud-run-upload`, and `cloud-run-transcoder`. Everything in `src/main.rs`'s `mod tests` — including the ~200 lines added here — is compile-checked but never executed. I confirmed `cargo test --locked` fails to link locally for exactly the reason in your PR body. Not this PR's problem, but "Validation: cargo check --tests" is weaker evidence than it looks, and it's a growing blind spot. Separate issue?

**5. Minor: error fields written for non-video blobs.**

In the new-blob path, `transcode_error_code` / `transcode_error_message` / `transcode_last_attempt_at` are assigned unconditionally, while `transcode_status` and `transcode_attempt_count` are gated on `is_video_mime_type`. The reset helpers are mime-gated, so this is inconsistent. Harmless today since the upload service only classifies transcodable types, but it can persist an error code on a blob whose status is `None`.

**6. Minor: `.unwrap()` decoupled from its guard.**

```rust
) {
    let error_code = transcode_error_code.as_ref().unwrap();
```

Safe only because the helper checks `is_some()` — a future edit to the helper turns this into a panic. `if let Some(error_code) = transcode_error_code.as_ref().filter(|_| should_record_...)` keeps them together.

**7. Scope: `Retry-After` in `upload_exposed_headers`.**

Unrelated to derivative failures and unmentioned in the PR description. It's a correct change, just flagging it for the description.

Approving on the strength of the fix once you've looked over a71ddde.

dcadenas (approved):
Approved after reviewer takeover.

The review identified five related state-consistency hazards in deduplicated and resumable uploads: a reupload could downgrade an existing terminal verdict, overwrite a derivative that was already Processing, treat derivative outcomes that resumable completion had not evaluated as clean, update the full metadata record from a cached value, and discard a failed metadata write while returning success.

The branch now contains two focused reviewer follow-ups. Rabble pushed `a71ddde`, which keeps terminal verdicts sticky and leaves in-flight derivative work authoritative. I then pushed `7a8f337`, which distinguishes evaluated derivative outcomes from unavailable ones, resets failures only after an evaluated clean result, reads authoritative metadata before updating an existing record, and propagates persistence failures. The combined regression coverage preserves the in-flight protections while covering direct and resumable observation decisions.

Validation included an isolated local service exercise on the final head. With terminal transcode and transcript failures seeded, resumable completion returned `200` while both HLS and subtitle endpoints remained `422` before and after completion. Test-target compilation and static analysis pass with existing warnings, all 115 reusable core tests pass, and all required pull-request checks are green.

Remaining uncertainty is limited to existing repository validation gaps: the complete host-linked test binary cannot run outside the edge runtime because its platform symbols are unavailable, although all test targets compile, and the repository-wide formatting check reports pre-existing drift outside the takeover diff.

@NotThatKindOfDrLiz please review both takeover commits and merge when ready, or follow up with any questions.

PR #162, opened 2026-07-30
fix(harness): cap ProofMode headers so the debugger stops reproducing the bug

### LINKED ISSUE #166: Track ProofMode debug harness header fidelity
## Summary

The debug upload harness should mirror the mobile client's ProofMode header behavior closely enough to reproduce upload failures and validate fixes.

## Scope

- Preserve the raw ProofMode manifest JSON when building the manifest header, instead of making header-budget decisions from a reserialized object.
- Keep an operator escape hatch for probing edge header limits with uncapped ProofMode headers.
- Run the harness test suite in CI so future header-shaping regressions are caught.

## Validation

- Harness unit tests cover raw manifest encoding, capped header behavior, and uncapped replay mode.
- CI runs the root scripts test suite.

hm21:
## Summary

- Cap the `X-ProofMode-*` headers the debug upload harness sends at 8 KiB combined, ordered cheapest-and-most-identifying first (`X-ProofMode-C2PA` → `-Signature` → `-PublicKey` → `-Attestation` → `-Manifest`)
- Mirror the new `X-ProofMode-PublicKey` header the client now sends, so the harness keeps reproducing what the app actually puts on the wire
- Warn on stderr naming what was dropped, so request-header size becomes visible instead of silent
- No server behaviour changes — this is the harness only

## Motivation

Replaying a real mobile proof through the harness reproduced the exact failure it exists to diagnose.

An Android device attestation is the `X509Certificate.toString()` dump of the whole cert chain — device-dependent in size — and it travels twice: base64-encoded in `X-ProofMode-Manifest` and again in `X-ProofMode-Attestation`. Measured against `media.divine.video`, combined request headers must stay under **128 KiB over HTTP/1.1** (past it the edge answers `502`) and under **64 KiB over HTTP/2** (past it the connection is closed mid-request). The HTTP/1.1 boundary is sharp: 130,800 bytes still reaches the handler (`401`), 131,000 returns `502`.

In practice that means a raw attestation over roughly 47.5 KB breaks the request (47,500 B → `401`, 48,000 B → `502`), so this bites on some devices and not others — a Pixel 8 Pro emits 53,826 bytes and always fails.

So anyone reaching for this harness to investigate a failing upload got an opaque `502` that reads like a server fault, when the request never had a chance. That is how the mobile-side bug stayed unexplained.

Worth recording while it is fresh: **nothing in this repo reads `X-ProofMode-*`.** The edge service has no reference to `proofmode` outside this script and docs, and `divine-upload-server` has none at all. The canonical manifest is the `proofmode` tag on the kind 34236 event, and the C2PA manifest is embedded in the uploaded bytes — which is what `cloud-functions/process-blob` reads. Dropping an oversized header therefore loses nothing.

The client-side fix is divinevideo/divine-mobile#6507; the 8 KiB budget here mirrors `BlossomUploadService.maxProofModeHeaderBytes` there.

## Related Issue

- Closes #166

## Validation

- [x] `python3 -m unittest scripts.tests.test_debug_upload_harness` — 16 tests pass, including the pre-existing `test_build_proof_headers_maps_expected_fields`, which pins backward compatibility: every header that shipped before keeps its exact value. The change is additive rather than byte-identical — a proof carrying `publicKey` now also sends `X-ProofMode-PublicKey` even when everything fits the budget; the pre-existing fixture has no `publicKey`, which is why its headers come out unchanged
- [ ] `cargo check --tests --locked` — not run, no Rust touched
- [ ] `cargo test --manifest-path cloud-run-upload/Cargo.toml --locked` — not run, no Rust touched
- [ ] `cargo clippy --locked --all-targets --all-features` — not run, no Rust touched

The Rust gates are skipped deliberately: this PR touches only the Python harness, its tests, and CI test discovery. Happy to run them if the repo prefers them green on every PR regardless of scope.

## Manual Test Plan / Notes

Not exercised end-to-end against production with a real proof JSON — the new path is covered by unit tests only here. The client-side twin of this logic has since run on a real Samsung Galaxy (divinevideo/divine-mobile#6507): the budget fired as designed — `sent 1121 bytes and dropped X-ProofMode-Attestation (11956B), X-ProofMode-Manifest (13512B)` — and the upload completed. One finding worth carrying over: that capture's manifest contained no `publicKey` at all (libProofMode emitted no `<hash>-pubkey.asc`), so a genuine Android proof of that shape sends the signature alone and `X-ProofMode-PublicKey` never fires — the signature/key pairing only holds for proofs that actually include the key. A maintainer replaying `--proof-json` with a big-attestation manifest should see the warning line and a completing request rather than a 502.

## Visuals / API Examples

- [x] No visual change

Warning emitted for an oversized proof:

```
warning: ProofMode headers over the 8192 byte budget; sending 3958 bytes and dropping X-ProofMode-Attestation (71768B), X-ProofMode-Manifest (75856B)
```

`X-ProofMode-C2PA` (60 B), `-Signature` (1,096 B) and `-PublicKey` (2,732 B) survive — 3,958 of the 8,192 byte budget.

## Public Artifact Review

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,6 +42,9 @@ jobs:
         working-directory: cloud-run-transcoder
         run: python3 -m unittest discover -s tests -p 'test_*.py'
 
+      - name: Run script Python tests
+        run: python3 -m unittest discover -s scripts/tests -p 'test_*.py'
+
       - name: Run clippy
         run: cargo clippy --locked --all-targets --all-features
 
--- a/scripts/debug_upload_harness.py
+++ b/scripts/debug_upload_harness.py
@@ -19,6 +19,20 @@ import urllib.request
 DEFAULT_TIMEOUT_SECONDS = 30
 SUCCESS_STATUSES = {200, 201, 204}
 
+# Combined wire budget for the X-ProofMode-* request headers, mirroring
+# BlossomUploadService.maxProofModeHeaderBytes in divine-mobile.
+#
+# media.divine.video rejects requests whose combined headers exceed 128 KiB
+# over HTTP/1.1 (with a 502) or 64 KiB over HTTP/2 (by closing the connection).
+# An Android hardware attestation chain is ~54 KB and lands in the manifest
+# twice — ~75 KB base64 in X-ProofMode-Manifest plus ~72 KB in
+# X-ProofMode-Attestation — so replaying a real proof through this harness
+# reproduced the client bug instead of diagnosing it.
+#
+# No Blossom server reads these headers; the canonical manifest is the
+# proofmode tag on the kind 34236 event.
+MAX_PROOF_HEADER_BYTES = 8192
+
 
 class HarnessError(RuntimeError):
     """Raised when the harness cannot execute a request or parse a response."""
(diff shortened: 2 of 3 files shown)


NotThatKindOfDrLiz:
@hm21 I took over the remaining harness follow-ups and pushed d0c863b.

Changes made:
- preserved the raw ProofMode manifest JSON from --proof-json for X-ProofMode-Manifest, so replay/header-budget decisions match the client instead of a Python reserialization
- added --max-proof-header-bytes, with 0 meaning uncapped, so the harness can still probe oversized-header edge limits
- added CI coverage for scripts/tests and expanded the harness regressions to 16 tests
- replaced Closes #none with Closes #166

Verification:
- python3 -m unittest scripts.tests.test_debug_upload_harness
- python3 -m unittest discover -s scripts/tests -p 'test_*.py'
- git diff --check origin/main...HEAD
- GitHub checks are green: test, semantic-pull-request, CLA

Please review the takeover commit and merge when you are ready.

NotThatKindOfDrLiz (approved):
Approved after takeover. The previous blockers are addressed: raw manifest JSON is preserved for replay fidelity, the cap can be disabled for edge-limit probing, scripts/tests now run in CI, and the PR closes #166. Local targeted tests and GitHub checks are green.

NotThatKindOfDrLiz:
## Review: Three blockers before merge

The PR faithfully mirrors divine-mobile#6507 and all 12 unit tests pass. However, three issues make the harness less faithful to the client than it claims:

### 1. ⚠️ Manifest re-serialization creates budget divergence

**Issue:** `build_proof_headers` re-encodes the manifest with `json.dumps(proof, separators=(",", ":"))`, losing the original wire-byte count. For captures with non-ASCII content (device labels, EXIF), the harness's drop decisions diverge from the client's.

**Measured example:**
- Client base64s original: attestation 1,627 B + manifest 4,676 B = 6,303 B → both fit, both sent
- Harness re-serializes: manifest inflates to 6,848 B → manifest DROPPED (different outcome)

**Why it matters:** The PR's entire justification is "keeps reproducing what the app actually puts on the wire." For oversized proofs, it now doesn't.

**Fix:** Have `load_proof_json` preserve the raw text and base64 that instead of re-encoding:
```python
def load_proof_json(path: str | Path) -> tuple[dict[str, object], str]:
    proof_path = Path(path)
    raw_text = proof_path.read_text(encoding="utf-8")
    return json.loads(raw_text), raw_text
```
Then in `build_proof_headers`, use the preserved raw text for the manifest value.

### 2. ⚠️ No way to disable or raise the cap from CLI

**Issue:** `max_bytes` has no `--max-proof-header-bytes` flag. The harness can no longer probe what happens with uncapped headers — exactly the failure mode the PR diagnosed.

**Why it matters:** An operator needing to verify an edge-side limit change, or re-run the 502 reproduction from the PR description, must edit the source.

**Fix:** Add to `build_parser()`:
```python
parser.add_argument(
    "--max-proof-header-bytes",
    type=int,
    default=MAX_PROOF_HEADER_BYTES,
    help="Cap for combined X-ProofMode-* header size (0 = unlimited)",
)
```
Then pass `max_bytes=args.max_proof_header_bytes` to `build_proof_headers`.

### 3. 🔴 CI doesn't run the harness tests

**Issue:** `.github/workflows/ci.yml` runs `python3 -m unittest discover -s tests -p 'test_*.py'` in `cloud-run-transcoder/`, never touching `scripts/tests/`. The three new harness tests are unenforced.

**Impact:** The `test` check passed on this PR without executing a single line of the new code. The PR's safety argument rests on those tests.

**Fix:** Add to `.github/workflows/ci.yml` under the test job:
```yaml
- name: Run harness tests
  run: python3 -m unittest scripts.tests.test_debug_upload_harness
```

### Minor: Link an issue

`Closes #none` should link a real issue. Recommend creating one (or using an existing tracking issue if one exists for the ProofMode budget work).

---

**None of these are large changes.** #1 and #3 are each ~5 lines; #2 is ~8 lines. The debug harness has no production blast radius, so if you prefer to merge with these as tracked follow-ups, that's defensible — but #3 should really be done regardless, since the tests added here are untested without it.

Also: this PR needs a human reviewer assigned per the repo's PR guardrails. I'll stop here pending your direction.

PR #165, opened 2026-07-31
perf(upload): drop the avoidable ffmpeg work from PUT /upload

### LINKED ISSUE #164: task(upload): drop the synchronous ffmpeg work from PUT /upload
## Description

`PUT /upload` runs ffmpeg and ffprobe synchronously for every video before it answers the client. We would like to propose removing at least the thumbnail half of it, and we have a question before anyone acts on that — see the bottom.

## What the handler does today

Per video upload, before the response is written (`cloud-run-upload/src/main.rs`, `cloud-run-upload/src/thumbnail.rs`):

1. The whole video — already buffered in memory as `all_bytes` — is written to a temp file.
2. `ffmpeg -y -ss 1 -i <video> -vframes 1 -vf scale=640:360:force_original_aspect_ratio=decrease -q:v 2 -f image2 <out.jpg>`
3. If that produced nothing, **ffmpeg runs a second time** with `-ss 0`.
4. The JPEG is read back and uploaded to GCS as `{hash}.jpg` — a further awaited round-trip.
5. The whole video is written to a temp file **a second time**, and `ffprobe -show_streams -select_streams v:0` reads width, height and rotation.

So: two full disk writes of the same bytes, one to two ffmpeg invocations, one ffprobe invocation, and one GCS upload, all on the client's response path.

## Three things we noticed

**The seek point does not match the product.** `-ss 1` seeks to one second. Divine is short-form; for anything shorter the first invocation yields nothing and the retry at `-ss 0` runs. A large share of uploads probably pays for ffmpeg twice.

**The ffmpeg call blocks an async worker.** `thumbnail.rs` uses `std::process::Command::output()` and is called directly from an `async fn` with no `spawn_blocking`, so a Tokio worker is blocked for the entire ffmpeg run. `probe_video_dimensions` right next to it correctly uses `tokio::process`. On Cloud Run with few threads, concurrent uploads can starve each other on this.

**The thumbnail is almost always discarded.** The mobile client extracts its own thumbnail at the frame the user picked as the cover — 210 ms by default, not 1 s — uploads it, and then explicitly prefers it. From `UploadManager.createSuccessfulUpload`:

> Prefer the custom cover we uploaded from the user's chosen frame (already stored on the upload) over the server's derived poster — otherwise a picked cover is silently replaced by `{videoHash}.jpg`.

The server's poster is not quite dead: the client accepts it when its own extraction or upload failed, so it is a fallback. But the residual it covers is narrow, because the client is already defensive on its own. `VideoThumbnailService.extractThumbnail` runs a four-step ladder — the chosen frame, the same frame again after 100 ms, then 50 ms, then duration/2 — and the upload that follows has its own three attempts with backoff.

And when the fallback does fire, the outcome is arguably worse than failing: the video is published with an arbitrary one-second frame instead of the cover the user chose, silently. Nobody is told.

## Why this reached us

We instrumented the mobile publish pipeline (divinevideo/divine-mobile#6554) and measured nine publishes on a real device. The client-side waste is fixed there — publishing is ~32% faster — and what remains is server-bound:

- The video upload uses roughly 10% of a 50 Mbit/s line.
- `PUT /upload` for a **16 572 byte** thumbnail took 2 273 to 7 216 ms across runs, with no relation to bandwidth. That path skips the ffmpeg branch entirely (`is_video_type` is false for images), so it is not explained by this issue — but it suggests the handler has fixed costs worth looking at in general.

## Our proposal

Remove the server-side thumbnail extraction and upload from the request path, and let mobile own the thumbnail outright.

The client already generates one on every upload, with the retry ladder above, and it produces a better frame than the server can — the one the user actually chose. We are willing to own the residual failure case: if our extraction and upload both exhaust their retries, the publish should fail visibly rather than quietly ship a frame the user never picked.

For the dimension probe, `spawn_blocking` is not needed since it already uses `tokio::process`, but reusing the bytes already on disk from step 1 would avoid the second write.

## The question

**Does anything other than the mobile client consume `{hash}.jpg`?** Web, feed rendering, OG images, moderation tooling — if any of those rely on the server-derived poster existing for every video, removing it is not free and we would rather know before proposing it further.

## Related

- divinevideo/divine-mobile#6554 — the client-side work and the measurements
- #163 — orphaned blobs from publishes that never completed

hm21:
## Summary

`PUT /upload` ran ffmpeg two to four times per video before answering the client — a full remux, one or two thumbnail extractions, and ffprobe. This removes two of those from the request path and leaves the third running only when it can change something.

**Thumbnail extraction is gone from the upload handler.** Clients supply their own cover — mobile extracts the frame the user picked and then explicitly prefers it over the server's — so the server poster was computed for every upload and discarded for nearly all of them. It is not lost: `GET /thumbnail/:hash` already downloads the blob, generates the poster, stores it at `{hash}.jpg` and returns it, and the edge already proxies a `{hash}.jpg` miss there. The poster is now produced on first request instead of on every upload. The response still returns the CDN URL, which resolves through that path, so no caller sees a new `null`.

**The remux is now conditional.** A video whose `moov` box already precedes `mdat` is web-compatible as-is; remuxing rewrites it into the layout it already has. A new top-level box scan (`cloud-run-upload/src/mp4.rs`) answers that in microseconds without spawning a process. divine-mobile exports with `shouldOptimizeForNetworkUse` on every render path, so its uploads skip the remux entirely.

## Motivation

Measured on a real device in divinevideo/divine-mobile#6554: the client-side publish waste is fixed there (~32% faster), and what remained was server-bound. The analysis on #164 established that nothing outside the mobile client depends on the server poster existing eagerly — divine-web and divine-funnelcake resolve the thumbnail from the event tags and never derive it from the video hash, and the moderation dashboard derives it but already degrades on a miss.

Two findings shaped the design:

- **The resumable upload path never generated a poster at all** (`handle_resumable_complete` returned `thumbnail_url: None`). Clients already cope with a missing server poster, which is what made the lazy poster safe to rely on. Review then aligned the two paths in the other direction: resumable completes now advertise the same lazily-resolving URL as `PUT /upload`, through a shared `video_thumbnail_url` helper.
- **The remux doubles as a validity gate** — `sanitize_error` is terminal in `classify_invalid_media_signal` and skips HLS transcoding and transcript generation. Skipping it had to preserve that. It does: `probe_error` from the ffprobe call is terminal in the same function and still runs on every video.

## Safety of the skip

**The remux output never leaves the process.** It is not stored, and the transcoder receives only `{hash, owner}` and fetches the original blob from GCS itself. So downstream services already worked on the unsanitized original before this change, and a skipped remux does not alter what any of them receive. Since the thumbnail extraction is gone, ffprobe is now the only consumer of the remuxed bytes.

That narrows what the skip can actually cost to two mirror-image cases. First: a file that ffmpeg cannot remux, that is nonetheless faststart, and that ffprobe reads without complaint — previously the remux failure would have flagged it `invalid_media`, now the consequence is a transcode attempt that may fail, reported by the transcoder, not silent corruption. Second, the inverse: a faststart file whose atoms make ffprobe fail but that the remux would have repaired — previously cleaned and transcoded, now flagged `invalid_media`. Both require a muxer that deliberately wrote faststart and still produced a file ffmpeg tooling stumbles over; everything else stays on the remux path.

The scan is deliberately conservative about which files reach that path. It returns `Some(true)` only when `moov` is positively found before `mdat`; non-MP4 containers, truncated files and unparseable input return `None` and still get remuxed. Every box must cover at least its own header — 16 bytes in the large-size form — so a malformed size can never rewind the scan onto bytes the file controls. And files with `moov` first come, in practice, from a muxer that was explicitly told to write faststart — our own renderer and transcoder. Camera captures from iPhone and Android put `moov` at the end, so they never take the skip.

## Also in this diff

- `extract_thumbnail` now runs under `spawn_blocking` in the on-demand handler, bounded by a four-slot semaphore. It shells out with a synchronous `Command`, which mattered little when one upload triggered it, but matters now that a cold feed can request many posters at once — and `spawn_blocking` alone would hold one ffmpeg process per request in flight, up to the blocking pool's size.
- The lazy poster path now coalesces on-demand generation per hash on each Cloud Run instance, bounds both same-hash waits and global ffmpeg-slot waits, and caches extraction failures briefly so bad videos do not retry full download plus ffmpeg on every request.
- `HEAD /{hash}.jpg` now follows the same lazy generation fallback as `GET /{hash}.jpg`, so advertised poster URLs do not 404 until a GET warms them.
- `POST /upload/resumable/:id/complete` now returns the same lazily-resolving poster URL as `PUT /upload` instead of `thumbnail_url: null`; both resolve through the same edge path.
- `classify_invalid_media_signal` loses the thumbnail-error parameter it already ignored — no caller can supply one anymore. The test that pinned "thumbnail failures are not terminal" went with it, since it could no longer fail; two tests covering reachable branches replace it.
- The uploaded bytes are held as `Bytes`, so the remaining clones are refcount bumps. A skipped remux now keeps one copy of the video in memory instead of three.

## Validation

- `cargo test --manifest-path cloud-run-upload/Cargo.toml --locked` — 34 passed, 15 of them new (10 for the box scan, 4 for the sanitize gate, 1 for the poster URL helper).
- `cargo check --tests --locked` — passed; existing dead-code warnings remain in the edge crate.
- The undersized-large-size regression test was verified against the pre-fix scanner: the crafted input misparsed to a fake `Some(true)` before the bound and returns `None` after it.
- `cargo clippy --manifest-path cloud-run-upload/Cargo.toml --locked --all-targets` — clean apart from a pre-existing `too_many_arguments` warning on `report_derivative_failure_to_sentry`, untouched here.
- `git diff --check` — clean.
- `rustfmt --edition 2021 --check src/main.rs cloud-run-upload/src/main.rs` still reports pre-existing formatting deviations reached through `mod` declarations and older spots in `src/main.rs`; unrelated rustfmt churn was not included.
- The box scan was checked against real ffmpeg output: a `+faststart` export classifies `Some(true)`, a default export `Some(false)`.

**Not yet run:** no end-to-end upload against the local stack or a deployed environment. Worth checking before merge that a mobile-rendered MP4 logs `Video is already faststart, skipping the derivative remux` and writes no `{hash}.jpg`, and that a subsequent fetch of `{hash}.jpg` generates it through the edge.

No visual change; this is a server-side request-path change.

## Related

- #164 — the analysis this implements
- divinevideo/divine-mobile#6554 — the client-side work and the measurements
- The remaining synchronous remux for non-faststart uploads is a separate question — it gates transcoding and needs its own design pass rather than a removal.

Closes #164

### DIFF
--- a/cloud-run-upload/src/main.rs
+++ b/cloud-run-upload/src/main.rs
@@ -34,10 +34,11 @@ use serde::{Deserialize, Serialize};
 use sha2::{Digest, Sha256};
 use std::{
     env,
-    sync::Arc,
+    sync::{Arc, LazyLock},
     time::{SystemTime, UNIX_EPOCH},
 };
 use tempfile::NamedTempFile;
+use tokio::sync::Semaphore;
 use tower::Service;
 use tower_http::cors::{Any, CorsLayer};
 use tracing::{error, info, warn};
@@ -965,6 +966,17 @@ async fn download_best_available_media_bytes(
 
 /// On-demand thumbnail generation endpoint
 /// Downloads video from GCS, generates thumbnail, stores it, returns the image
+/// Concurrent on-demand poster extractions allowed per instance.
+///
+/// Each extraction runs a whole ffmpeg process over a temp copy of the video.
+/// `spawn_blocking` alone would let a cold-feed burst launch one per request
+/// up to the blocking pool's size; this bound keeps such a burst from
+/// exhausting the instance's CPU and memory while queued requests wait.
+const MAX_CONCURRENT_THUMBNAIL_EXTRACTIONS: usize = 4;
+
+static THUMBNAIL_EXTRACTION_SLOTS: LazyLock<Semaphore> =
+    LazyLock::new(|| Semaphore::new(MAX_CONCURRENT_THUMBNAIL_EXTRACTIONS));
+
 async fn handle_thumbnail_generate(
     State(state): State<Arc<AppState>>,
     Path(hash): Path<String>,
@@ -1053,6 +1065,11 @@ async fn handle_thumbnail_generate(
     // Generate thumbnail. `extract_thumbnail` shells out with a synchronous
     // `Command`, so it must not run on a runtime worker — a cold feed can ask
     // for many posters at once and would otherwise stall unrelated requests.
+    // The semaphore bounds how many of those ffmpeg processes run at once.
+    let _permit = THUMBNAIL_EXTRACTION_SLOTS
+        .acquire()
+        .await
+        .expect("the extraction semaphore is never closed");
     let extraction = tokio::task::spawn_blocking(move || thumbnail::extract_thumbnail(&video_data))
         .await
         .map_err(|e| anyhow!("Thumbnail extraction task panicked: {}", e))
--- a/cloud-run-upload/src/main.rs
+++ b/cloud-run-upload/src/main.rs
@@ -823,7 +823,9 @@ async fn stream_to_gcs_with_hash(
     // `Bytes` keeps the later clones refcount bumps instead of full copies.
     let original_bytes = Bytes::from(original_bytes);
 
-    // Derivative generation (probe/transcode) can use sanitized bytes.
+    // The remux output stays in-process: only ffprobe reads it, and it is never
+    // stored. The transcoder fetches the original blob from GCS itself, so a
+    // skipped remux changes nothing about what downstream services receive.
     let mut sanitize_error = None;
 
     let derivative_bytes = if needs_derivative_sanitize(content_type, &original_bytes) {
--- a/cloud-run-upload/src/main.rs
+++ b/cloud-run-upload/src/main.rs
@@ -1,6 +1,7 @@
 // ABOUTME: Rust Cloud Run service for Blossom blob uploads
 // ABOUTME: Handles Nostr auth validation, streaming upload to GCS, and SHA-256 hashing
 
+mod mp4;
 mod resumable;
 mod thumbnail;
 
@@ -643,31 +644,14 @@ async fn process_upload(
     )
     .await?;
 
-    let mut thumbnail_error: Option<String> = None;
-    // Extract thumbnail for videos (non-blocking - failures don't fail the upload)
-    let thumbnail_url = if thumbnail::is_video_type(&content_type) {
-        match extract_and_upload_thumbnail(
-            &state.gcs_client,
-            &state.config.gcs_bucket,
-            &state.config.cdn_base_url,
-            &sha256_hash,
-            &all_bytes,
-        )
-        .await
-        {
-            Ok(url) => {
-                info!("Generated thumbnail for {}", sha256_hash);
-                Some(url)
-            }
-            Err(e) => {
-                error!("Thumbnail extraction failed for {}: {}", sha256_hash, e);
-                thumbnail_error = Some(e.to_string());
-                None
-            }
-        }
-    } else {
-        None
-    };
+    // The poster is no longer extracted here. Clients supply their own cover
+    // (mobile uploads the frame the user picked and prefers it over ours), and
+    // a request for `{hash}.jpg` that finds nothing in GCS is served by
+    // `GET /thumbnail/:hash` via the edge. Returning the URL keeps it resolvable
+    // for callers that use it; it is generated on first request instead of on
+    // every upload.
+    let thumbnail_url = thumbnail::is_video_type(&content_type)
+        .then(|| format!("{}/{}.jpg", state.config.cdn_base_url, sha256_hash));
 
     let mut probe_error: Option<String> = None;
     // Probe video dimensions (non-blocking - failures don't fail the upload)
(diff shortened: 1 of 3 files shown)


NotThatKindOfDrLiz:
@hm21 I took over the last blocking pieces and pushed the branch back on top of current `main`.

What changed:
- Added per-hash single-flight around lazy poster generation on each Cloud Run instance.
- Added bounded waits for both same-hash generation and the global ffmpeg extraction semaphore.
- Added a short negative cache for videos whose poster extraction fails, so bad videos do not retry full download + ffmpeg on every request.
- Made `HEAD /{hash}.jpg` use the same lazy generation fallback as `GET /{hash}.jpg`.
- Preserved Cloud Run `422` and busy `503` outcomes through the edge thumbnail proxy.
- Updated the PR body and rebased onto current `main`.

Validation:
- `cargo test --manifest-path cloud-run-upload/Cargo.toml --locked` passed, 34 tests.
- `cargo check --tests --locked` passed, with existing edge dead-code warnings.
- `cargo clippy --manifest-path cloud-run-upload/Cargo.toml --locked --all-targets` passed with the existing `too_many_arguments` warning.
- `git diff --check` passed.
- GitHub CI is green on `1d883bb`.

Please review the takeover changes and merge when you are ready.

NotThatKindOfDrLiz (approved):
Final pass is clean after the takeover changes. The lazy poster contract now has HEAD parity, bounded/coalesced generation, short negative caching for extraction failures, the branch is rebased onto current main, and CI is green.

PR #177, opened 2026-08-07
docs(cdn): add vendor capability notes and CDN steering design

rabble:
## Summary

CDN and delivery-origin evaluation: vendor capability notes, a steering design, a measurement tool with tests, and the measurement records produced with it.

- `docs/cdn-evaluation-status.md` — entry point; current state, results, open questions
- `docs/cdn-object-storage-vendor-notes.md` — verified vendor limits and pricing (Fastly, R2, B2, bunny, GCS)
- `docs/superpowers/specs/2026-08-06-cdn-delivery-steering-design.md` — server-side steering between two CDNs
- `scripts/probe_cdn_delivery.py` + tests — the measurement tool (41 tests)
- `docs/measurements/` — five records, including one superseded and explicitly marked

## Findings that constrain the architecture

- **Fastly Object Storage's public S3 endpoint is capped at ~150 Mbps / 100 req/s per bucket**, exempt only for Fastly's own network. A bucket used as the delivery replica cannot then serve a second CDN.
- **No published durability figure or availability SLA** for Fastly Object Storage; no object lock, versioning, or lifecycle rules, so legal hold must stay on the authoritative store.
- **"Zero egress" covers Object Storage into Compute and Deliver only.** CDN-to-viewer and shield-to-edge bandwidth are billed; reads are still Class B operations.
- **Blossom does not require single-origin delivery** — BUD-02 defines the descriptor `url` as any publicly accessible endpoint, and all outbound URLs already funnel through `to_descriptor()`.
- **No streaming miss is configured anywhere**, so every cache miss buffers the whole object before the client sees a byte. This interacts with stripping `Range` on cache fill.

## Access control: the replica is the ACL

Replicating only on moderation-approval means a gated blob is never in the replica, so a second CDN cannot serve what it can never fetch. No tokens, no deny-list, no edge compute, no per-request cost. **Validated end to end on real infrastructure** — content Fastly serves right now returns 404 on bunny purely because the replicator did not put it there.

The design is asymmetric and the docs say so: never-replicated content is protected structurally; already-replicated content is an ordinary distributed delete.

## Takedown findings

- Deleting from the replica does **not** take content down — the CDN kept serving from cache until purged.
- **`b2_hide_file` does not remove data.** After hiding, the upload version was still listed and still downloadable by file id. Only `b2_delete_file_version` on every version removes it. This is how the B2 UI delete and an S3 `DeleteObject` against a versioned bucket behave.
- Per-record purge works on both CDNs keyed on the SHA-256 — Fastly by surrogate key (already implemented), bunny by `{hash}*` URL wildcard (verified).

## Measurement results

In North America — ~85% of delivered watch time — Fastly, bunny Volume, and bunny Standard are within 1 ms of each other. Volume's weakness is confined to regions with no PoP, which is under 1% of traffic.

| Region | Share | Fastly | bunny Volume | bunny Standard |
|---|---:|---:|---:|---:|
| us-central1 | ~85% | 13 ms | 14 ms | 13 ms |
| europe-west1 | ~11% | 3 ms | 9 ms | 11 ms |
| australia-southeast1 | <1% | 2 ms | 97 ms | 3 ms |

## Review feedback addressed

Responding to @NotThatKindOfDrLiz:

1. **Committed JSON contradicted the method claim** — fixed. The probe's cache-status selection now prefers the serving CDN header (their commit), the Wellington measurement was regenerated, and the JSON records `hit: 45` on all three edges.
2. **Timing metric was ambiguous** — fixed and it changed conclusions. The probe now keeps a keep-alive connection per edge and reports handshake cost separately from response latency, with `--no-reuse` for cold-start. **The earlier headline that bunny Volume beats Fastly in NA/EU is retracted** — it was an artifact of counting a ~296 ms trans-Pacific TLS handshake as server latency. All four regions were re-measured.
3. **PR description** — this rewrite.

Their reviewer-takeover commit is included and its cache-header fix was a real bug catch.

## Scope and validation

Documentation, one script, and its tests. No production code, config, or behaviour changes.

- 41 tests pass (`python3 -m unittest discover -s scripts/tests -p 'test_*.py'`)
- Vendor claims quoted from primary sources, linked, verified 2026-08-06
- Traffic figures and commercial analysis deliberately excluded — this repository is public; that half is in `divine-context`

## Known gaps, recorded not hidden

- Every US number comes from a datacentre; a US consumer-connection datapoint is missing and covers 85% of traffic.
- The verdict logic manufactures failures at low latency (3 ms vs 2 ms reads as +43.5%); needs an absolute floor.
- Cache hit ratio on a real library is unmeasured and determines the bill.
- Temporary probe zones and buckets are live; teardown tracked in #178.


NotThatKindOfDrLiz (changes requested):
@rabble I pushed a small reviewer takeover commit for the bounded local fixes:

- recorded all cache-related probe headers and changed the selected cache status to prefer the serving CDN header (`cdn-cache`) before forwarded origin `x-cache`; added regression tests
- indexed the new CDN/vendor and measurement docs in `docs/README.md`
- corrected stale design-doc references (`types.rs:387`, `src/main.rs:5945`, `scripts/probe_cdn_delivery.py`)
- marked `is_publicly_cacheable` as proposed helper text and made the bucket pseudocode short-input safe
- opened #178 and referenced it from the temporary probe-zone teardown TODO

Validation passed locally with `python3 -m unittest discover -s scripts/tests -p 'test_*.py'` (36 tests), and the pushed GitHub checks are green.

Requesting changes for the remaining author-owned work before merge:

1. Rerun the Wellington measurement with the updated probe, or otherwise regenerate the raw JSON and prose so the cache HIT/MISS claim is supported by the committed data. The current JSON still says both bunny edges were `miss: 45`, while the measurement doc says every response was a cache HIT.
2. Decide what the timing metric should mean. Either measure with connection reuse, or rename/frame the current `ttfb_ms` as a cold connect-plus-response-header metric and adjust the playback-startup conclusions accordingly.
3. Refresh the PR description so it covers the probe/measurement commits and the remaining validation story.

Please review the reviewer commit along with the remeasurement updates when you make them. I’m intentionally not approving while the committed measurement artifact still contradicts the method claim.

rabble:
@NotThatKindOfDrLiz thanks — all three points addressed, and point 2 changed the conclusions.

**1. JSON contradicted the method claim.** Fixed. Your `select_cache_status` change was a real bug catch: the original read `x-cache` first, which on a bunny edge is Fastly's *forwarded* header rather than bunny's own `cdn-cache`. The Wellington measurement has been regenerated and the committed JSON now records `hit: 45` on all three edges, matching the prose.

**2. Timing metric.** You were right and it mattered more than a naming fix. `ttfb_ms` bundled DNS + TCP + TLS + response headers with no connection reuse, so a ~296 ms Wellington→Los Angeles handshake was charged to server latency on *every* request.

The probe now holds a keep-alive connection per edge, samples handshake cost separately, and reports both; `--no-reuse` restores cold-start behaviour. Five regression tests cover the split, including that a slow handshake must not fail an edge on response time.

All four regions were re-measured. **The headline finding is retracted:** "bunny Volume beats Fastly in NA/EU" was an artifact of the old method. Corrected, Volume *ties* Fastly in the US and *loses* in Europe. `2026-08-07-four-region-summary.md` now carries a SUPERSEDED banner pointing at `2026-08-07-four-region-corrected.md`.

The recommendation survives for a different reason: weighted by watch time, North America is ~85% of delivered bytes and all three edges sit within 1 ms there, so Volume is indistinguishable at half the Standard rate. The earlier analysis over-weighted an Oceania result — Australia is 0.87% of watch time, and Wellington, where the measurements were taken, is where the team sits rather than a user population. Both are now labelled as such.

**3. PR description.** Rewritten to cover the probe and measurement commits and the validation story.

Two things your review surfaced that I've recorded but *not* fixed, so they don't get lost:

- The verdict logic manufactures failures at low latency — Sydney Standard reads as FAIL at +43.5%, which is 3 ms against 2 ms. A percentage margin needs an absolute floor.
- Every US number comes from a GCE instance. That's 85% of traffic measured under datacentre conditions, and a consumer-connection datapoint is the most valuable outstanding measurement.

Merging now at @rabble's direction rather than waiting on re-approval — flagging that explicitly since you said you were intentionally withholding it. Happy to take follow-ups in a new PR.

PR #184, opened 2026-08-07
docs(cdn): add bunny delivery rollout implementation plan

rabble:
## Summary

Turns the validated CDN evaluation (#177) into production wiring: `docs/superpowers/plans/2026-08-07-bunny-delivery-rollout-plan.md`.

## The wiring is smaller than expected

Two chokepoints cover every path, so no call-site hunting:

| Concern | Hook | Existing call sites |
|---|---|---|
| Replicate on approval | `metadata::update_blob_status(hash, Active)` | `admin.rs:1040`, `main.rs:3101` |
| Purge + delete | `purge_edge_cache(hash)` | 10, incl. `delete_policy.rs`, GDPR vanish, tombstone |

Deletion already fans out through one function. Extending it reaches creator delete, admin moderation, GDPR erasure, tombstone and legal hold.

## Never emit a vendor hostname

An earlier revision had `to_descriptor` emit a bunny hostname and argued that because Nostr events embed URLs at publish time, existing content simply wouldn't migrate — framed as a safe gradual rollout. **That was wrong**, and the plan says so explicitly. Baking a vendor hostname into published immutable events permanently surrenders server-side control: no fast failover, no way to apply a renegotiated contract to the back catalogue, and weaker vendor leverage.

The fix is to emit hostnames we own and route below the URL, so CDN choice becomes a DNS change that applies retroactively to already-published content.

## Layered source selection

Content addressing supplies the enabling property: the client can verify what it got by hashing it, so multi-source fetching is safe in a way ordinary CDN failover isn't.

1. `v.divine.video`, DNS-steered — the floor, no client work
2. Named per-CDN hostnames, all owned and independently retargetable — no client work
3. Client-side preference + failover — **sketched, deliberately deferred**
4. Blossom kind-10063 server lists — multi-source fallback for third-party clients, free

Layer 3 is held back on evidence, not caution: the corrected measurements put all three CDNs within 1 ms in North America, ~85% of traffic. Its failure modes are recorded — the preference document becomes a playback-path dependency, argmax selection herds clients and wrecks forecasting against committed volume, and cost attribution becomes emergent.

## Moderation safety

- **Replicate on approval, never on upload** — a correctness requirement, since the replica *is* the access control.
- **Ordering trap recorded:** the descriptor must not route to the replica until replication is confirmed, or a freshly approved video 404s for its first viewers.
- **`purge_edge_cache` must fail loudly first.** It currently returns early with an `eprintln!` when the token is missing and its doc comment says it never fails the calling request. Correct for a cache, wrong for a moderation control.

## Federation boundary

Blobs on another Blossom server are fetched directly by the client and are not our bandwidth; selection applies only to hostnames we own. Flags an unverified possible inversion — fallback backends for `cdn.satellite.earth` and `image.nostr.build` are declared in `fastly.toml.example` with no usage found in `src/`. If that chain proxies rather than redirects, we pay to serve third-party content. Recorded as a question, not a finding.

## Testing

Integration step 3 is the one that matters: **reject a blob, assert it never reaches the replica** — that catches a replicate-on-upload regression, the failure that silently voids the whole design. Plus a continuous canary asserting a tombstoned hash 404s everywhere, `CacheErrorResponses == False`, and **zero versions** in the replica rather than absence by name.

## Staged rollout

Stage 1 runs replication in production with nothing depending on it, so drift surfaces before any user is affected. Rollback at every stage is a config flip or a DNS change.

## Scope

Documentation only. No code, config, or behaviour changes.

## Not resolved

- Cache hit ratio on the real library is unmeasured and determines the bill.
- Every US latency number came from a datacentre; 85% of traffic has no consumer-connection datapoint.
- Blocked on DNS for `v.divine.video` before end-to-end failover can be tested.

### DIFF
--- a/docs/cdn-evaluation-status.md
+++ b/docs/cdn-evaluation-status.md
@@ -16,6 +16,7 @@ Entry point for the delivery-origin and second-CDN evaluation. Start here, then 
 | [measurements/2026-08-07-replica-as-acl-validation.md](measurements/2026-08-07-replica-as-acl-validation.md) | Working test system proving absence-is-denial |
 | [measurements/2026-08-07-takedown-drill.md](measurements/2026-08-07-takedown-drill.md) | Takedown semantics — **B2 hide does not erase data** |
 | [measurements/2026-08-07-nz-wellington.md](measurements/2026-08-07-nz-wellington.md) | First measurement; superseded in scope, still valid for Oceania |
+| [superpowers/plans/2026-08-07-bunny-delivery-rollout-plan.md](superpowers/plans/2026-08-07-bunny-delivery-rollout-plan.md) | **Implementation plan** — hook points, client migration, tests, staged rollout |
 | [../scripts/probe_cdn_delivery.py](../scripts/probe_cdn_delivery.py) | The measurement tool |
 
 Commercial analysis — traffic volumes, cost models, vendor questions — is deliberately **not** in
--- a/docs/superpowers/plans/2026-08-07-bunny-delivery-rollout-plan.md
+++ b/docs/superpowers/plans/2026-08-07-bunny-delivery-rollout-plan.md
@@ -96,6 +96,75 @@ must never receive a `v.divine.video` URL at all. DNS decides *which CDN*; the s
 *whether the content is eligible to leave the Fastly path*. Those are separate concerns and both are
 required.
 
+## Delivery source selection — layered
+
+Content addressing supplies a property most CDN setups lack: **the client can verify what it got.**
+Fetch from any source, hash it, compare to the requested SHA-256. You cannot be served wrong bytes
+without detecting it. That makes multi-source fetching safe in a way ordinary CDN failover is not,
+and it is the foundation for everything below.
+
+### Layer 1 — `v.divine.video`, DNS-steered
+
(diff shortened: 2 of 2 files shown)


rabble (commented):
Reviewed in depth. Every code citation in the plan checks out against the branch: both `update_blob_status(hash, Active)` sites (`admin.rs:1040`, `main.rs:3101`), all ten `purge_edge_cache` call sites, `to_descriptor` (`blossom-core/src/types.rs:387`), `get_base_url` (`main.rs:6051`), and the full URL-to-storage mapping table against `delete_blob_gcs_artifacts`. The four fallback backends in `fastly.toml.example` are indeed unreferenced in `src/` — the flagged open question is accurate.

One finding, fixed on the branch (17a7fae): the plan said Pending blobs served publicly is documented in `vcl/fetch.vcl`. Neither the repository snippet nor the live version-14 fetch snippet mentions Pending; the behaviour lives in `BlobMetadata::access_for` (`Pending` → `Allowed`). The citation now points there. Substance of the plan is unaffected.

The self-correcting section on vendor hostnames is the best part of this doc — owning the delivery hostname is the right call, and the honest accounting of why beats a polished wrong answer.

PR #185, opened 2026-08-07
fix(scripts): require an absolute floor before failing an edge on latency

rabble:
## Summary

The probe's verdict logic compared p95 response latency by percentage alone. Against a single-digit millisecond baseline that produces false failures — Sydney measured Fastly at 2 ms and bunny Standard at 3 ms, and the tool reported **FAIL at +43.5%**.

One millisecond is imperceptible. A tool that cries wolf there gets ignored when it matters.

## Change

A regression must now exceed **both** the relative margin and an absolute floor (default 10 ms, `--floor-ms`). `--floor-ms 0` restores the old behaviour.

Re-judged with the floor:

| Edge | Verdict | Reason |
|---|---|---|
| bunny Standard | **PASS** | +50.0% but only +1.0 ms, within the 10 ms floor |
| bunny Volume | **FAIL** | +4750.0% (+95.0 ms), outside both |

Verdict reasons now carry the absolute delta alongside the percentage, so a result can be sanity-checked without returning to the raw JSON.

## Deliberately unchanged

- **Errors bypass both gates.** An edge that drops requests fails however fast the survivors were.
- **All-errors stays `NO_DATA`, not `FAIL`.** An edge that could not be measured has not been shown to be worse. A test asserting this is included, because I initially wrote the errors test wrong and it caught me.

## Validation

77 tests pass (`python3 -m unittest discover -s scripts/tests -p 'test_*.py'`). Six new tests cover the floor: small absolute delta passes despite a large percentage, large delta still fails, delta exactly at the floor passes, errors are not rescued, all-errors is NO_DATA, and a zero floor restores pure-percentage behaviour.

Closes the caveat recorded in `docs/measurements/2026-08-07-four-region-corrected.md`.

### DIFF
--- a/docs/measurements/2026-08-07-four-region-corrected.md
+++ b/docs/measurements/2026-08-07-four-region-corrected.md
@@ -38,7 +38,7 @@ deleted), Wellington from a consumer connection. All cache hits.
 |---|---:|---:|---:|---:|
 | **us-central1** | **~85%** | 13 ms | **14 ms** | **13 ms** |
 | europe-west1 | ~11% | **3 ms** | 9 ms | 11 ms |
-| australia-southeast1 | <1% | 2 ms | **97 ms** | 3 ms |
+| australia-southeast1 | <1% | 2 ms | **97 ms** | 3 ms ✅ |
 | nz-wellington † | **0%** | 12 ms | **144 ms** | 18 ms |
 
 **† Wellington is where the team is based, not where users are.** New Zealand does not appear in the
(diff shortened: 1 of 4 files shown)


rabble (commented):
Verified locally: 77 tests pass (`python3 -m unittest discover -s scripts/tests`), including all six new `TestLowLatencyFloor` cases. The two-gate logic is right: margin checked first, floor only rescues a margin breach, errors bypass both, all-errors stays `NO_DATA`, `--floor-ms 0` restores pure-percentage behaviour. The re-judged Sydney table in the measurements doc matches the code's arithmetic (+1.0 ms passes, +95.0 ms fails).

Also verified against the live Fastly service while cross-checking: the teardown note about pull zone 6289364 carrying `v.divine.video` is consistent with what the rollout plan in #184 assumes.

No findings. The errors-bypass-floor test and the NO_DATA distinction are exactly the cases that would have regressed silently.

PR #186, opened 2026-08-07
fix(cache): stop marking mutable derivatives as immutable

rabble:
## Summary

Stop marking repairable media derivatives as browser-immutable while retaining the long, purgeable edge TTL.

| Content | `Cache-Control` | `Surrogate-Control` |
|---|---|---|
| Blob | `public, max-age=31536000, immutable` | `max-age=31536000` |
| Derivative | `public, max-age=86400` | `max-age=31536000` |

The seven HLS, rendition, and transcript GET/HEAD call sites now use the derivative policy. The immutable blob policy is unchanged.

The VCL fetch snippet now preserves Compute's explicit browser `Cache-Control` instead of replacing every successful response with a one-year immutable policy. It still enforces a one-year Fastly edge TTL and uses the immutable policy as a fallback when the origin supplies no browser policy.

The repository VCL also now records the already-live `beresp.do_stream = true` block so applying this snippet cannot regress the production streaming fix, and records the live `vcl_miss` Range-strip snippet it depends on as `vcl/miss.vcl` (byte-identical to the active snippet).

## Motivation

`/backfill-fmp4` rebuilds renditions and `scan_and_repair_vtts.py` rewrites transcripts while their URLs remain unchanged. A browser that received `immutable` could retain a broken derivative for a year, and browser caches cannot be purged.

The one-day browser TTL bounds recovery time for previously cached clients. The edge TTL and per-hash `Surrogate-Key` remain unchanged, so normal edge invalidation still works immediately.

This was found while validating the second delivery-CDN plan and is tracked by #184.

## Validation

- `cargo test -p blossom-core --locked` — 119 passed, including four cache-policy tests executed by CI
- `cargo check --tests --locked` — passed (pre-existing warnings only)
- `cargo clippy --locked --all-targets --all-features` — passed (pre-existing warnings only; the warning introduced by the original PR head was removed)
- Active VCL snippet inspected before the change; the live `do_stream` behavior is preserved verbatim
- Combined merge simulation with #184, #185, and #193 completed without conflicts

## Manual deployment and verification

CI publishes the Compute service only. After merge:

1. Clone the active VCL service version.
2. Apply this repository's `vcl/fetch.vcl` as the versioned fetch snippet.
3. Validate and activate the cloned VCL version.
4. Purge both Fastly services.
5. Verify a derivative returns `public, max-age=86400` without `immutable`, a blob remains immutable for one year, range streaming still works, and the full smoke suite passes.

Rollback is activation of the previous VCL version plus the previous Compute version.

## Not addressed here

The repair script still does not issue a purge after rewriting a VTT. This change bounds the browser-cache impact to one day; it does not add that purge. The follow-up remains in #184.

### DIFF
--- a/blossom-core/src/cache_policy.rs
+++ b/blossom-core/src/cache_policy.rs
@@ -1,0 +1,71 @@
+// ABOUTME: Shared cache policy for immutable blobs and repairable derivatives.
+// ABOUTME: Keeps browser and edge cache lifetimes testable outside Fastly Compute.
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct PublicCacheHeaders<'a> {
+    pub cache_control: &'static str,
+    pub surrogate_control: &'static str,
+    pub surrogate_key: &'a str,
+}
+
+pub fn immutable_blob_cache_headers(hash: &str) -> PublicCacheHeaders<'_> {
+    PublicCacheHeaders {
+        cache_control: "public, max-age=31536000, immutable",
+        surrogate_control: "max-age=31536000",
+        surrogate_key: hash,
+    }
+}
+
+pub fn mutable_derivative_cache_headers(hash: &str) -> PublicCacheHeaders<'_> {
+    PublicCacheHeaders {
+        cache_control: "public, max-age=86400",
+        surrogate_control: "max-age=31536000",
+        surrogate_key: hash,
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::{immutable_blob_cache_headers, mutable_derivative_cache_headers};
+
+    fn max_age(cache_control: &str) -> u64 {
+        cache_control
+            .split("max-age=")
+            .nth(1)
+            .and_then(|rest| rest.split(|c: char| !c.is_ascii_digit()).next())
+            .and_then(|digits| digits.parse().ok())
+            .expect("Cache-Control must carry a numeric max-age")
+    }
+
+    #[test]
+    fn blobs_are_immutable_for_a_year() {
+        let headers = immutable_blob_cache_headers("abc123");
+
+        assert_eq!(headers.cache_control, "public, max-age=31536000, immutable");
+        assert_eq!(headers.surrogate_control, "max-age=31536000");
+        assert_eq!(headers.surrogate_key, "abc123");
+    }
+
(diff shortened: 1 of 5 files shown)


rabble (commented):
Verified end to end.

**Tests run in CI, as the description claims.** The four cache-policy tests live in `blossom-core/src/cache_policy.rs`, which CI executes via `cargo test -p blossom-core --locked`. Local run: 119 passed, matching the description. `cargo check --tests --locked` and clippy also pass.

**Call-site audit.** All seven HLS/rendition/transcript GET+HEAD sites switched; the remaining `add_cache_headers` sites are the raw-blob GET/HEAD (correct — content-addressed) and thumbnails/audio. Thumbnails and derived audio keep the immutable policy; I traced the mutation paths and neither has an in-place rewrite today (audio bytes only change via out-of-band deletion followed by re-extraction), so the line the PR draws matches the PR's evidence. Worth a sentence in the doc trail, not a code change.

**The VCL change fixes a live bug beyond its stated scope.** The active version-14 fetch snippet still stamps `public, max-age=31536000, immutable` on every 200/206 — including responses Compute marked `private, no-store` for Restricted/AgeRestricted content. Preserving the origin's explicit policy repairs that. The recorded `do_stream` block matches the live snippet verbatim, and the referenced live `vcl_miss` Range-strip snippet exists and matches the comment's description.

**One adjacent gap (not introduced here, flagging rather than changing per deployment-config rules):** the live `vcl_miss` snippet (`unset bereq.http.Range;`, snippet 'Cache full objects for range requests') is not recorded in `vcl/`. The fetch comment references it, but someone applying repo snippets to a fresh VCL version would silently lose the Range strip. Recommend recording it as `vcl/miss.vcl` — either here with maintainer sign-off or in a follow-up.

Merge-order note for whoever deploys: CI auto-deploys Compute on merge, but until the VCL snippet is applied the old fetch snippet keeps stamping immutable over Compute's new derivative headers — the merge is inert, not harmful. The body's manual VCL steps cover it.

rabble:
Follow-up to my review: pushed e5df1de recording the live `vcl_miss` Range-strip snippet as `vcl/miss.vcl` (byte-identical to the active snippet 'Cache full objects for range requests' on VCL service v14), authorized by the maintainer. PR body updated to match.

PR #187, opened 2026-08-08
feat(compiler): add internal compilation service and editorial workspace

rabble:
## Summary

Adds the internal compilation service end to end: a Cloud Run renderer (`cloud-run-compiler/`), an editorial web workspace (`compiler-web/`, served at `compiler.divine.video`), and the release workflow and edge wiring that deploy them.

- **`cloud-run-compiler/`** — Rust service that takes a signed kind 30005 Nostr list, resolves each referenced video event to its media source, normalizes clips, burns creator credits and Divine branding, and renders one external compilation per requested aspect (9:16, 1:1, 16:9). Jobs are asynchronous: create, poll, fetch. State lives in Firestore; outputs upload through `upload.divine.video`.
- **`compiler-web/`** — React workspace to open one of your signed lists, reorder clips, choose per-clip framing, save the reordered list back to the relay, and queue a render. Nothing it does publishes to Divine.
- **List resolution** accepts both tag shapes: the Divine app writes `e` tags holding video event ids, the compiler editor writes `a` tags holding addressable coordinates. Both load, in literal signed order.
- **Timeline clips show the creator's profile name** (kind 0 `display_name`/`name`), falling back to a short npub, resolved in one batched relay query after the clips render.
- **Release** through `.github/workflows/compiler-release.yml` and the managed infrastructure path in the runbooks.

## Motivation

Editors need to turn a curated Nostr list into a single shareable cut without hand-editing video or publishing intermediate artifacts to Divine. This is that path, kept entirely outside the normal publish flow.

## Review round 2 — what changed

Rebased onto current `main` (conflicts resolved), and every review finding addressed:

| Finding | Resolution |
| --- | --- |
| drawtext quoting unsafe for apostrophes | Credits now go through `textfile=`, which has no escaping surface. Added `apostrophe_bearing_credit_renders`, an ffmpeg-backed render of `O'Brien, 100%: \ [the] real`. **Verified the test is a real regression test**: reverted to the inline `text='...'` form in the container and it fails; with `textfile=` it passes. |
| Upload service contradiction | Resolved by removing the flag. Production `upload.divine.video` is the GKE-hosted `divinevideo/divine-upload-server`, which has no `generateDerivatives` field — I read its source to confirm. Its resumable completion path stores the object without starting thumbnail, HLS, or transcription work, so suppression comes from the endpoint, not a flag that would read as a guarantee production does not honour. `ResumableUploadInitRequest` no longer sends it, a test asserts its absence, and the deployment design and superseded plan task now say all of this. The dead `cloud-run-upload/` changes are dropped from the branch. |
| Service rejects tag shapes the editor filters out | `NostrEvent::list_slots()` keeps unsupported `a` coordinates and malformed `e` tags as `ListSlot::Unsupported`, and `resolve_sources` reports them through the existing dropped-clip path against their original index. A list that opens and saves in the UI no longer 400s at render. |
| New code not covered by PR CI | Added `compiler` and `compiler-web` CI jobs: compiler tests, compiler clippy with `-D warnings`, the Firestore emulator test, plus `npm ci` / `npm test` / `npm run build`. |
| Firestore test silently passes without the emulator | Now `#[ignore]`d with an explicit reason and it panics if run without `FIRESTORE_EMULATOR_HOST`. CI starts the emulator container and runs it, covering the rate-limit transaction, the claim CAS, and lease reclaim. |
| One relay filter per coordinate | Addressable lookups are grouped by author and kind with `#d` arrays and chunked (100 identifiers per filter, 10 filters per request); event-id and profile queries are chunked too. |
| Editor opens one `get` per coordinate | Same treatment client-side: `coordinateFilters()` groups and chunks, `videoEvents()` issues a handful of `querySync` calls and picks the newest event per coordinate. Unit-tested. |
| No lease/requeue for running jobs | Jobs carry `lease_expires_at`. `claim_next` prefers queued work and otherwise reclaims a running job whose lease has passed; the worker heartbeats every 60s while rendering and clears the lease on completion. Added a Firestore composite index for `status` + `lease_expires_at`. |
| Unrelated CDN view-counting changes | Split into #188, for the owner of that pipeline to review. Not in this branch any more. |

## Linked issue

No tracking issue exists for this work; design and rollout plans are in `docs/`. Happy to open one if the team wants it tracked separately.

## Validation

- `cargo test --manifest-path cloud-run-compiler/Cargo.toml --locked` — passes.
- `cargo clippy --manifest-path cloud-run-compiler/Cargo.toml --all-targets --all-features --locked -- -D warnings` — clean.
- `docker run divine-compiler-test:local cargo test --test render -- --ignored` — both ffmpeg-backed renders pass in the production-like container, including the apostrophe regression.
- `npm --prefix compiler-web test -- --run` — 29 tests pass. `npm --prefix compiler-web run build` — clean (`tsc --noEmit` + vite).
- Confirmed against `wss://relay.divine.video` that kind 0 profiles resolve for real clip authors, so timeline names populate.

Not run locally: the Firestore emulator test (CI runs it) and any production deploy from this branch.

## Screenshots

`docs/compiler-editor-empty.png` shows the empty editor state. The creator-name change replaces the truncated hex label under each timeline clip and in the preview footer with the author's profile name.


cloud-run-compiler/src/render.rs
```diff
@@ -0,0 +1,383 @@
...
+    font_path: &Path,
+) -> String {
+    if duration_ms == 0 {
+        return format!("[{input}]null[{output}]");
+    }
+    let end = start + f64::from(duration_ms) / 1000.0;
+    let fade_in_end = (start + 0.3).min(end);
+    let fade_out_start = (end - 0.3).max(start);
+    format!(
+        "[{input}]drawtext=fontfile={font}:text='{text}':\
```
NotThatKindOfDrLiz: This is still unsafe for creator names containing an apostrophe. `escape_drawtext()` emits `\'`, but this value is inside ffmpeg single quotes, where backslash does not escape the quote for the filter parser. A display name like `O'Brien` can terminate the `text='...'` token and break the whole filtergraph. Please switch this to a no-escaping surface such as `textfile=` or use ffmpeg's close/escape/reopen quoting, and add a real ffmpeg smoke/regression test with an apostrophe-bearing credit.

docs/superpowers/specs/2026-07-26-compilation-deployment-design.md
```diff
@@ -0,0 +1,144 @@
...
+- A release dispatch that proposes a pinned compiler image digest in
+  `divine-iac-coreconfig`.
+
+The application workflow may publish artifacts, but it does not imperatively
+create or mutate Cloud Run, IAM, Firestore, DNS, or Access resources.
+
+### `divine-upload-server`
+
+`upload.divine.video` is provided by the GKE-hosted
+`divinevideo/divine-upload-server`, not the historical `cloud-run-upload/`
```
NotThatKindOfDrLiz, line 50: This deployment design says production `upload.divine.video` is the GKE-hosted `divine-upload-server`, not this repo's historical `cloud-run-upload/`, and that no upload-server change is required. But the compiler relies on `generateDerivatives: false` being honored by the upload endpoint, while this PR only adds that plumbing to `cloud-run-upload/`. Please resolve the contradiction before merge: either update the actual production upload service, or remove this dead/incorrect service change and document why production already suppresses derivatives.

cloud-run-compiler/src/domain.rs
```diff
@@ -0,0 +1,370 @@
...
+    }
+
+    /// Ordered video references, accepting both the addressable `a` coordinates
+    /// written by the compiler editor and the `e` event ids written by the
+    /// Divine mobile app.
+    pub fn video_references(&self) -> Result<Vec<VideoReference>, ValidationError> {
+        self.tags
+            .iter()
+            .filter_map(|tag| match (tag.first().map(String::as_str), tag.get(1)) {
+                (Some("a"), Some(value)) => Some(Ok(value)
```
NotThatKindOfDrLiz: The service currently treats any `a` tag that is not kind 34235/34236, and any malformed `e` tag, as a request-failing validation error. The editor's `videoReferences()` filters those tags out instead, so a list can open/save in the UI and then 400 at render time. Please align the service with the editor: ignore/drop unsupported reference shapes and report them through the existing dropped-clip path rather than rejecting the entire list.

Cargo.toml
```diff
@@ -3,7 +3,7 @@
 
 [workspace]
 members = ["blossom-core"]
-exclude = ["cloud-run-upload", "cloud-run-transcoder"]
+exclude = ["cloud-run-upload", "cloud-run-transcoder", "cloud-run-compiler"]
```
NotThatKindOfDrLiz, line 6: Adding `cloud-run-compiler` to the workspace excludes means the existing root `cargo clippy --all-targets --all-features` CI step will not cover the new compiler. Since `.github/workflows/ci.yml` is not updated, the Rust compiler service and the React app can both break while PR checks stay green. Please add explicit PR CI steps for `cargo test/clippy --manifest-path cloud-run-compiler/Cargo.toml`, `npm --prefix compiler-web test -- --run`, and `npm --prefix compiler-web run build`.

cloud-run-compiler/tests/store.rs
```diff
@@ -0,0 +1,139 @@
...
+        .create(&job("same", "one@example.com", 20))
+        .await
+        .unwrap_err();
+
+    assert!(error.to_string().contains("already exists"));
+}
+
+#[tokio::test]
+async fn firestore_transaction_round_trip_when_emulator_is_available() {
+    if std::env::var("FIRESTORE_EMULATOR_HOST").is_err() {
```
NotThatKindOfDrLiz: Returning early here makes the Firestore integration test pass whenever the emulator is absent, and CI does not set `FIRESTORE_EMULATOR_HOST`. That leaves the durable job store, rate-limit transaction, and claim CAS path with no executed automated coverage. Please either run a Firestore emulator in CI for this test or mark it as an explicit ignored/manual test and add automated coverage for the persistence boundary that actually executes.

cloud-run-compiler/src/source.rs
```diff
@@ -0,0 +1,372 @@
...
+        }
+        client.connect().await;
+        Ok(Self { client, timeout })
+    }
+}
+
+#[async_trait]
+impl SourceRepository for RelaySourceRepository {
+    async fn addressable_events(&self, coordinates: &[String]) -> Result<Vec<NostrEvent>> {
+        let filters: Vec<Filter> = coordinates
```
NotThatKindOfDrLiz: This builds one relay filter per addressable coordinate in a single REQ. Large editor-authored lists can produce hundreds of filters, which common relays cap or reject; the failure then looks like every clip is missing. Please chunk this or batch by author/kind with `#d` arrays instead of sending one filter per clip.

compiler-web/src/nostr/relay.ts
```diff
@@ -0,0 +1,124 @@
...
+        this.relays,
+        { ids: eventIds },
+        { maxWait: 10_000 },
+      )
+      for (const event of events as NostrEvent[]) {
+        byId.set(event.id, event)
+      }
+    }
+
+    const pairs = await Promise.all(
```
NotThatKindOfDrLiz: This opens one relay `get` per addressable reference. A normal 64-clip list becomes 64 concurrent REQs on every open, while the event-id branch above is already batched. Please batch the coordinate lookup as well, or at least chunk requests to stay under relay limits.

cloud-run-compiler/src/store.rs
```diff
@@ -0,0 +1,492 @@
...
+            let Some(document) = rows.into_iter().find_map(|row| row.document) else {
+                self.rollback(&transaction).await;
+                return Ok(None);
+            };
+            let mut job = decode_job(&document)?;
+            if job.status != JobStatus::Queued {
+                self.rollback(&transaction).await;
+                continue;
+            }
+            job.status = JobStatus::Running;
```
NotThatKindOfDrLiz, line 426: Once a job is flipped to `Running`, there is no lease, heartbeat, timeout, or requeue path. If the Cloud Run instance is replaced or crashes mid-render, the job can stay running forever and the editor eventually gives up. If v1 intentionally ships without recovery, please track that accepted debt independently; otherwise add a lease/reclaim mechanism before merge.

vcl/log_cdn_views.vcl
```diff
@@ -12,20 +12,28 @@
...
-#   - Bare SHA256 paths (not thumbnails, HLS segments, VTT, etc.)
-#   - HTTP 200 responses (not 206 range requests, 404s, etc.)
+#   - Video byte responses (Content-Type: video/*)
+#   - Canonical and derivative video paths for the same SHA256
+#   - Any successful response that delivered video bytes
 #
 # Every logged row = one view. No dedup, no rate limiting.
 # No client IP stored — only POP for geographic distribution.
 
 if (req.method == "GET"
```
NotThatKindOfDrLiz, line 21: This CDN view-counting behavior is unrelated to the compiler/editor feature and changes production logging semantics. Please split the VCL and CDN view-counting docs/spec/plan changes into their own PR so they can be reviewed by the owner of that pipeline.

NotThatKindOfDrLiz (changes requested):
Thanks for putting this together. I’m requesting changes because there are still correctness and release-gate issues that can break normal editor/compiler use or let this merge without exercising the new code.

Blocking before merge: fix the ffmpeg drawtext quoting for apostrophes with an ffmpeg-backed test, resolve which upload service actually serves production and carries derivative suppression, align editor/service handling of non-video or malformed list tags, wire compiler and compiler-web tests/builds into PR CI, and split the unrelated CDN view-counting changes out of this PR. The PR also currently reports merge conflicts and has no linked issue/tracking item for accepted v1 debt.

Other items called out inline, such as Firestore coverage, relay batching, and job lease/requeue behavior, should either be fixed here or explicitly tracked before this ships.

cloud-run-compiler/src/render.rs
rabble: Fixed. Credits now go through `textfile=` instead of an inline `text='...'` token, so there is no escaping surface at all. Added `apostrophe_bearing_credit_renders`, an ffmpeg-backed render of `O'Brien, 100%: \ [the] real`, and verified it is a real regression test: reverting to the inline form in the container makes it fail, `textfile=` makes it pass.

docs/superpowers/specs/2026-07-26-compilation-deployment-design.md
rabble: Resolved by removing the flag. I read `divinevideo/divine-upload-server` — it has no `generateDerivatives` field, and its resumable completion path stores the finalized object without starting thumbnail, HLS, or transcription work. So suppression comes from the endpoint the compiler calls, and sending a flag would read as a guarantee production does not honour. `ResumableUploadInitRequest` no longer sends it, `tests/upload.rs` asserts its absence, and the deployment design plus the superseded plan task now spell this out. The dead `cloud-run-upload/` changes are dropped from the branch.

cloud-run-compiler/src/domain.rs
rabble: Fixed. `NostrEvent::list_slots()` keeps unsupported `a` coordinates and malformed `e` tags as `ListSlot::Unsupported` instead of failing, and `resolve_sources` reports them through the existing dropped-clip path against their original index, so the drop lines up with the position the editor shows. Covered by `drops_event_references_that_are_not_event_ids`, `drops_unsupported_addressable_kinds_without_failing_the_request`, and `unsupported_list_tags_are_dropped_clips_rather_than_a_failed_render`.

Cargo.toml
rabble: Added. CI has a `compiler` job (`cargo test` + `cargo clippy -- -D warnings` on the compiler manifest, plus the Firestore emulator test) and a `compiler-web` job (`npm ci`, `npm test -- --run`, `npm run build`).

cloud-run-compiler/tests/store.rs
rabble: Fixed. The test is now `#[ignore]`d with an explicit reason and panics if run without `FIRESTORE_EMULATOR_HOST`, so an absent emulator shows as unrun rather than green. CI starts the emulator container and runs it, and the test now covers the rate-limit transaction, the claim CAS, and lease reclaim end to end against a fresh collection.

cloud-run-compiler/src/source.rs
rabble: Fixed. Coordinates are grouped by author and kind into `#d` array filters and chunked at 100 identifiers per filter, 10 filters per REQ. The event-id and profile queries are chunked the same way, since both could also exceed relay array caps on a large list.

compiler-web/src/nostr/relay.ts
rabble: Fixed. `coordinateFilters()` groups coordinates by kind and author and chunks them, `videoEvents()` now issues a handful of `querySync` calls instead of one `get` per clip and picks the newest event per coordinate, and the profile lookup is chunked too. Unit-tested in `lists.test.ts`.

cloud-run-compiler/src/store.rs
rabble: Implemented rather than deferred. Jobs carry `lease_expires_at`; `claim_next` prefers queued work and otherwise reclaims a running job whose lease has passed, and the worker heartbeats every 60s while rendering and clears the lease on completion. Added the `status` + `lease_expires_at` composite index. Covered by memory-store tests and by the Firestore emulator test.

vcl/log_cdn_views.vcl
rabble: Split out into #188 so the owner of the view-counting pipeline can review it separately. It is no longer in this branch.

rabble:
One deployment follow-up the CI green does not cover: the new \`status\` + \`lease_expires_at\` composite index is declared in \`cloud-run-compiler/firestore.indexes.json\`, but production Firestore indexes are owned by the \`compiler-service\` Terragrunt stack in \`divine-iac-coreconfig\`, and the Firestore emulator does not enforce composite indexes — so the green emulator test is not evidence the reclaim query works in production.

Until that index is added to the coreconfig stack, real Firestore will return FAILED_PRECONDITION on the expired-lease query, which fires on every idle poll once the queued query comes back empty. Queued jobs still claim normally, so the visible symptom is only a warn log per poll and a reclaim path that never actually reclaims. Happy to open the coreconfig PR if that is the right sequencing.

.github/workflows/compiler-release.yml
```diff
@@ -0,0 +1,189 @@
...
+        default: true
+        type: boolean
+      deploy_worker:
+        description: Deploy the protected editor Worker after Cloud Run exists
+        required: true
+        default: false
+        type: boolean
+
+env:
+  COMPILER_IMAGE: us-central1-docker.pkg.dev/dv-platform-prod/containers-production/divine-compiler
```
NotThatKindOfDrLiz, line 18: This publishes to the same production image name used by the standalone `divinevideo/divine-compiler` release workflow. Coreconfig also names `divinevideo/divine-compiler` as the application repository for this stack, so merging this creates two repos that can publish/promote the same `divine-compiler` image. Please resolve repository ownership before adding this release path here.

cloud-run-compiler/src/store.rs
```diff
@@ -0,0 +1,585 @@
...
+        let document: FirestoreDocument = response
+            .error_for_status()
+            .context("Firestore rejected job read")?
+            .json()
+            .await
+            .context("decode Firestore job")?;
+        Ok(Some(decode_job(&document)?))
+    }
+
+    async fn save(&self, job: &Job) -> Result<()> {
```
NotThatKindOfDrLiz, line 376: `save()` is a blind full-document patch. Since `renew_lease()` reads and then calls this same save path, an in-flight heartbeat can land after the worker writes a terminal result and revert the job back to the old running document with a fresh lease. Terminal writes and lease renewals need fencing, such as an ownership token and/or Firestore updateTime preconditions.

cloud-run-compiler/src/store.rs
```diff
@@ -0,0 +1,585 @@
...
+                "currentDocument": { "updateTime": update_time }
+            });
+            if self.commit(&transaction, json!([write])).await? {
+                return Ok(Some(job));
+            }
+        }
+        Ok(None)
+    }
+
+    async fn renew_lease(&self, id: &str) -> Result<()> {
```
NotThatKindOfDrLiz, line 443: This lease renewal is not tied to the worker that claimed the job. If the lease expires while the original worker is still rendering, another worker can reclaim the job and both workers can later blind-write terminal state. Please add a fencing token or equivalent CAS check so stale workers cannot renew or complete jobs after reclaim.

docs/superpowers/specs/2026-07-26-compilation-deployment-design.md
```diff
@@ -0,0 +1,153 @@
+# Compilation Service Deployment: Design Spec
+
+**Status:** Approved
+**Date:** 2026-07-26
+**Site:** `https://compiler.divine.video`
+**Application repository:** `divinevideo/divine-blossom-stt`
```
NotThatKindOfDrLiz, line 6: This repository does not resolve on GitHub. These deployment docs are the operator source of truth, so they need to name the actual application repository once ownership is decided.

docs/superpowers/plans/2026-07-26-compilation-managed-deployment.md
```diff
@@ -0,0 +1,147 @@
+# Compilation Service Managed Deployment Implementation Plan
+
+> **For agentic workers:** Execute this plan with the test-driven-development
+> and verification-before-completion skills. Do not deploy production resources
+> while implementing the plan.
+
+**Goal:** Make `divine-iac-coreconfig` the authoritative owner of the private
+production compiler service and its internal edge, with image promotion from
+`divine-blossom-stt`.
```
NotThatKindOfDrLiz, line 9: Same ownership issue: this plan says image promotion comes from `divine-blossom-stt`, but that repo does not exist and coreconfig currently documents `divinevideo/divine-compiler` as the compiler app repo.

NotThatKindOfDrLiz (changes requested):
Requesting changes. The review-round fixes are real, but this PR should not merge into divine-blossom as-is. Coreconfig names divinevideo/divine-compiler as the compiler application repository, and that repo is live and ahead of this branch, so merging this creates two repos publishing/promoting the same production compiler image. If ownership is intentionally moving to divine-blossom, that needs an explicit maintainer decision plus coreconfig/runbook updates and a plan to carry over the fixes that currently exist only in divine-compiler.

There is also a correctness gap in the new lease implementation: terminal saves and heartbeats are not fenced, so a stale worker or in-flight heartbeat can overwrite a completed/reclaimed job. Please move the compiler work to the owning repo, fix the lease with a worker ownership token or CAS/preconditioned terminal writes, and correct the deployment docs before this ships.

dcadenas (changes requested):
The compilation service and editor here duplicate `divinevideo/divine-compiler`, which the production infrastructure config names as the compiler application repository.
Both that repository and the release workflow added on this branch build and promote the same production image, so merging this would leave two repositories publishing one image tag.
That needs resolving before this can go in.

Was reopening this branch intentional?
The earlier pull request from the same branch was closed on 26 July with a note that the compilation editor and renderer had moved to their own repository.
I could not find a recorded decision reversing that.
If ownership is meant to move back here, that is a maintainer call and would also need the infrastructure config and runbooks updated to match.

### Work here that should not be lost

Some of what this branch adds does not appear to exist in the compiler repository today.
Whichever way the ownership question lands, these are worth porting or tracking there:

- Job recovery — lease and heartbeat handling, so a job whose renderer disappears is reclaimed instead of sitting in a running state.
- Safer credit text rendering, which removes the quoting problems apostrophes cause.
- Tolerant handling of malformed or unsupported list entries, so a list that opens in the editor does not fail at render time.

I have not compared the two services at runtime, so treat that as a list to check rather than a settled gap.

rabble:
Closing because compiler service ownership remains in divinevideo/divine-compiler. Merging this would create a second implementation and deployment path publishing the same service image, while the lease-fencing concern raised in review still needs to be resolved in the owning repository. The valuable work here—job recovery and lease fencing, text rendering, and malformed-entry handling—should be ported as focused changes to divine-compiler instead of creating split ownership in Blossom.

PR #188, opened 2026-08-08
feat(cdn): count views from delivered video bytes

rabble:
## Summary

Split out of #187, where it was unrelated to the compiler/editor work.

Widens CDN view logging so a view is counted from delivered video bytes rather than only from a bare SHA-256 path with a `200`:

- Matches canonical and derivative video paths for the same hash (`/<sha>`, `.mp4`, `/720p`, `/480p`, HLS segments), with query strings tolerated.
- Requires `Content-Type: video/*`, a 2xx status, and `body_bytes_written > 0`, so a request that delivered no video bytes is not a view.
- Log rows gain `"v":2`, plus `path`, `status`, and `cache` (`fastly_info.state`); `sha256` is now extracted from the path rather than assumed to be the whole path.

## Motivation

Range requests (`206`) and derivative paths carried real playback that the previous `^/<sha>$` + `200` rule never counted, so view counts undercounted systematically. The `v` field lets the aggregation side tell old rows from new ones.

## Linked issue

None. The design and rollout notes are in `docs/superpowers/specs/2026-05-23-lossless-view-capture-design.md` and `docs/superpowers/plans/2026-05-23-lossless-view-capture.md`, both included here.

## Validation

- `scripts/tests/test_cdn_view_vcl.py` parses the regexes out of `vcl/log_cdn_views.vcl` and exercises them against real request paths; CI runs it. This version has still not been applied to the Fastly service.
- Reviewed against the existing snippet semantics only. **Needs a review from the owner of the view-counting pipeline before it ships**, since it changes both the counting rule and the log row shape that downstream aggregation parses.
- **Blocked on divine-funnelcake.** The raw-capture sink this depends on is not built: `000149` there is `000149_fix_view_counts_sort_key`, and `cdn_media_delivery_events` appears in no ref of that repository. The runbook now gates Fastly activation on that work landing.

No user-visible UI change.

## Downstream tracking

- divinevideo/divine-funnelcake#1070 tracks the required v2 sink and subscriber.
- This PR is non-activatable until that issue is complete, reviewed by the view-counting pipeline owner, deployed, and verified end to end. Merging this source-only VCL change does not authorize Fastly activation.


vcl/log_cdn_views.vcl
```diff
@@ -12,20 +12,28 @@
...
+#   - Canonical and derivative video paths for the same SHA256
+#   - Any successful response that delivered video bytes
 #
 # Every logged row = one view. No dedup, no rate limiting.
 # No client IP stored — only POP for geographic distribution.
 
 if (req.method == "GET"
-    && req.url ~ "^/[0-9a-fA-F]{64}$"
-    && resp.status == 200) {
+    && req.url ~ "^/[0-9a-fA-F]{64}($|\\?|\\.mp4(\\?|$)|/(720p|480p)(\\.mp4)?(\\?|$)|/hls/stream_(720p|480p)\\.(ts|mp4)(\\?|$))"
```
NotThatKindOfDrLiz: This pattern is over-escaped for Fastly VCL. VCL string literals do not consume backslashes as C-style escapes, so the regex engine sees double backslashes, not the intended single regex escapes. In practice the whitelist degenerates to matching any path beginning with /<64hex>, including routes the design says to exclude. Use single backslashes in the VCL string, and add a smoke test that validates the actual VCL literal rather than a separately typed Ruby regex.

vcl/log_cdn_views.vcl
```diff
@@ -12,20 +12,28 @@
...
+    && resp.http.Content-Type ~ "^video/"
+    && resp.status >= 200
+    && resp.status < 300
+    && resp.body_bytes_written > 0) {
   log {"syslog "} req.service_id {" cdn-view-logs :: "}
     {"{"}
+      {""v":2,"}
       {""ts":"} time.start.sec {","}
-      {""sha256":""} regsub(req.url, "^/", "") {"","}
+      {""sha256":""} regsub(req.url, "^/([0-9a-fA-F]{64}).*", "\1") {"","}
```
NotThatKindOfDrLiz: Please normalize the extracted hash with std.tolower(). The condition accepts uppercase hex, but the service normalizes hashes server-side and downstream joins expect lowercase values; uppercase requests would produce rows that miss the join.

vcl/log_cdn_views.vcl
```diff
@@ -12,20 +12,28 @@
...
+      {""path":""} regsub(req.url, "\\?.*$", "") {"","}
```
NotThatKindOfDrLiz: This has the same VCL escaping issue and will strip the entire URL instead of only the query string. With the current literal, the pattern can match at offset 0 and replace the whole path with an empty string, so v2 rows lose the path that downstream classification depends on.

docs/runbooks/cdn-view-counting.md
```diff
@@ -51,10 +51,23 @@ In the Fastly dashboard for VCL service `ML7R82HKfmTaqTpHExIDVN`:
...
    - **Topic:** `cdn-view-logs`
    - **Secret key:** paste contents of `fastly-pubsub-key.json`
-3. Add a VCL snippet:
-   - **Type:** `log` (vcl_log subroutine)
-   - **Content:** paste from `vcl/log_cdn_views.vcl`
-4. Activate the new version
+   - **Format version:** `2` (`vcl_log`)
+   - **Format:**
+     ```text
+     {"v":2,"ts":%{time.start.sec}V,"sha256":"%{regsub(req.url, "^/([0-9a-fA-F]{64}).*", "\1")}V","path":"%{regsub(req.url, "\\?.*$", "")}V","status":%{resp.status}V,"bytes":%{resp.body_bytes_written}V,"pop":"%{server.datacenter}V","cache":"%{fastly_info.state}V"}
```
NotThatKindOfDrLiz: The production endpoint format duplicates the broken escaping from the snippet. Please fix this copy too, or make one source clearly authoritative and keep the runbook subordinate so these regexes cannot drift. Also validate that the endpoint format compiles in a cloned Fastly version before activation.

docs/runbooks/cdn-view-counting.md
```diff
@@ -51,10 +51,23 @@ In the Fastly dashboard for VCL service `ML7R82HKfmTaqTpHExIDVN`:
...
+3. Create or update the response condition named `cdn-view-log-condition`:
+   ```vcl
+   req.method == "GET"
+   && req.url ~ "^/[0-9a-fA-F]{64}($|\\?|\\.mp4(\\?|$)|/(720p|480p)(\\.mp4)?(\\?|$)|/hls/stream_(720p|480p)\\.(ts|mp4)(\\?|$))"
+   && resp.http.Content-Type ~ "^video/"
+   && resp.status >= 200
+   && resp.status < 300
+   && resp.body_bytes_written > 0
+   ```
+4. Activate the new version.
```
NotThatKindOfDrLiz: This activates Fastly before running migration 000149 and deploying the v2-capable subscriber below. That contradicts the design doc rollout order and risks sending v2/range/derivative rows to the current v1 pipeline. Reorder the runbook so the Funnelcake migration and subscriber are live before Fastly activation.

NotThatKindOfDrLiz (changes requested):
Requesting changes. The current Fastly VCL/logging config does not implement the intended path policy: the doubled backslashes are preserved by VCL string syntax, so the regexes behave differently from the Ruby smoke test. That can emit empty path values and over-broadly capture any /64hex... video route. The rollout order also needs to put the Funnelcake migration/subscriber ahead of Fastly activation before this is safe to ship.

vcl/log_cdn_views.vcl
dcadenas: Confirmed, and fixed in 092c332. Fastly's own regex example uses a single backslash — `req.url ~ "^/products(?:/?.*)\z"` — and string literals take percent escapes, not backslash escapes. So `\\?` was an optional literal backslash, which matches the empty string and let the whole alternation accept any path under `/<sha>`.

The smoke test you asked for is `scripts/tests/test_cdn_view_vcl.py`. It parses the literal out of this file rather than retyping it, which is the only version that can catch this. Reverting the file to 773ce23 fails 9 of its 12 tests.

vcl/log_cdn_views.vcl
dcadenas: Fixed in df678f2, on both `sha256` and `path`. Uppercase is genuinely reachable: `parse_hash_from_path` validates with `is_ascii_hexdigit` and then returns `to_lowercase()`, so an uppercase request is served video bytes normally and would have logged a hash that joins against nothing.

vcl/log_cdn_views.vcl
dcadenas: Confirmed, and it was the most damaging one — `path` would have been empty on every row, which takes out `media_path_type` classification downstream.

Fixed in 092c332, then hardened in 3ebb59b to `"\?[\s\S]*$"`. A dot does not cross a newline, so `\?.*$` left the tail of `/<sha>?a=1\n"x` in the field, and `path` is interpolated into the JSON row unescaped.

docs/runbooks/cdn-view-counting.md
dcadenas: Fixed in 092c332, and the drift problem is now enforced rather than requested: `vcl/log_cdn_views.vcl` is authoritative, and `test_cdn_view_vcl.py` fails if the copies here diverge from it.

On compiling before activation — agreed, and step 4 now requires saving into a cloned version and confirming it first, with the reason stated: the format string nests braces and quotes inside `%{...}V`, and a parse failure there is invisible from outside.

docs/runbooks/cdn-view-counting.md
dcadenas: Reordered in 405f29f. Investigating it turned up more than an ordering problem: `000149` on divine-funnelcake main is `000149_fix_view_counts_sort_key`, and `cdn_media_delivery_events` appears in no ref of that repository. The raw-capture sink is unbuilt, so the ordering alone would not have made activation safe.

7339271 and 216b57f replace the prerequisite with a blocker that says so and gates activation on that work landing.

dcadenas (approved):
Widening the counting rule is the right call, and the design doc's raw-first tradeoff — overcapture at the edge, decide policy in ClickHouse — is the part worth keeping.

The regexes could not have worked as written. VCL string literals take percent escapes, not backslash escapes, so the doubled backslashes reached the regex engine as literal characters. Every one of the escaping comments on this PR was correct.

### Changes I pushed

- Single-backslash escapes in the snippet, the runbook, and the design doc.
- `std.tolower` on the logged hash and path, since uppercase hex is served and normalised server-side.
- The query strip now consumes past a newline, so nothing from a query string can land unescaped in the JSON row.
- Rollout reordered, and the runbook now carries a blocker rather than a prerequisite.
- A test that reads the regexes out of the VCL file and runs them, plus a check that the runbook copies match.

The blocker is the one that changes what this PR means. The raw-capture sink it feeds is not built: on divine-funnelcake main `000149` is `000149_fix_view_counts_sort_key`, and `cdn_media_delivery_events` appears in no ref of that repository. The ordering fix alone would not have made activation safe, so activation is now gated on that work landing.

### Checks

- `python3 -m unittest discover -s scripts/tests`: 83 tests, 12 of them new. Reverting the VCL to the previous head fails 9 of the 12.
- `cargo check --tests --locked`, `cargo clippy --locked --all-targets --all-features`
- `cargo test -p blossom-core --locked`, plus the upload and transcoder suites
- `python3 -m unittest discover -s tests` in `cloud-run-transcoder`: 23 tests

None of that speaks for the Fastly service itself — nothing here is Compute code. The endpoint format still has to be saved into a cloned version and confirmed before activation, which the runbook now says.

### Follow-ups

- The ClickHouse raw table and a v2-capable subscriber, in `divine-funnelcake`. Everything here is inert until they exist.
- The implementation plan still carries `[x]` checkboxes over that funnelcake migration. I left the author's record alone and named the file in the runbook blocker instead.

NotThatKindOfDrLiz:
@rabble I pushed a scoped cleanup commit (`8bb9b39`) to the PR branch.

What changed:
- widened the CDN view allow-list to include the video extensions this service already serves as `video/*` (`m4v`, `webm`, `mov`, `mkv`, `ogv`, `avi`) while keeping the `Content-Type: video/*` guard;
- added regression cases for those extensions to `scripts/tests/test_cdn_view_vcl.py`;
- cleaned the new plan/spec docs so they no longer point at `divine-blossom-stt`, no longer include personal absolute paths, and no longer mark the unbuilt funnelcake subscriber/migration work as complete;
- marked the design-doc response condition as illustrative, with `vcl/log_cdn_views.vcl` plus the runbook as the drift-tested authoritative copies.

Local verification: `python3 -m unittest discover -s scripts/tests` passes: 83 tests.

Please review the takeover commit when you get a chance. I am still requesting changes for the remaining rollout blocker: this PR must not merge as activatable until the `divine-funnelcake` v2 sink/subscriber work exists, is reviewed by the view-counting pipeline owner, and is tracked with an issue or equivalent explicit follow-up.

NotThatKindOfDrLiz (changes requested):
Requesting changes for the remaining work outside the cleanup commit.

The local PR-introduced cleanup issues are addressed by `8bb9b39`, and the prior VCL escaping/query-strip/drift-test objections are fixed. The remaining blocker is rollout correctness: the `divine-funnelcake` sink this emits to still does not exist, and the PR body still has no linked tracking issue despite the activation blocker.

Before merge, please either:

- land/link the `divine-funnelcake` work that creates `cdn_media_delivery_events` and updates `cdn-view-subscriber` to parse `"v":2`, with review from the view-counting pipeline owner; or
- keep this PR explicitly non-activatable, link a tracking issue for that sink/subscriber work, and make the PR body/runbook unambiguous that Fastly activation is blocked until that issue is done.

Do not activate this on Fastly until the downstream v2 sink exists and has been verified.

rabble:
Fresh dependency check: the Funnelcake repository still has no cdn_media_delivery_events v2 sink/table, and its current cdn-view-subscriber parser still consumes only the v1 {ts, sha256, bytes, pop} envelope. This PR must remain blocked and must not be activated until that downstream sink is implemented, deployed, and verified end to end. The existing change request therefore still stands; please link the downstream tracking issue before returning this for review.

rabble:
The remaining rollout blocker now has an owner and durable dependency: divinevideo/divine-funnelcake#1070 tracks the v2 raw sink, backward-compatible subscriber, production-scale schema review, deployment, and end-to-end verification.

I updated the runbook and PR description to make this branch explicitly non-activatable until #1070 is complete. Commit `180c50f`; 83 script tests pass and `git diff --check` is clean. Nothing was activated on Fastly. Please re-review the current head under the source-only/non-activatable option from the outstanding change request.

NotThatKindOfDrLiz:
@rabble I pushed `0f8eb69` to finish the takeover fix. It makes the SHA-256 extractor consume newline-delimited suffixes before interpolating the hash into JSON, keeps the runbook format synchronized, and adds a regression test that fails against the prior `.*` pattern.

Validation on the final head:
- `python3 -m unittest scripts.tests.test_cdn_view_vcl.TestFieldExtraction.test_extracted_sha256_is_json_safe`
- `python3 -m unittest discover -s scripts/tests` — 84 tests passed
- `git diff --check`
- GitHub `test`, semantic-title, and CLA checks passed

The full PR diff is clean and the source change is ready to merge. Please review the takeover commit and merge when you are ready. Fastly activation remains blocked on divinevideo/divine-funnelcake#1070; merging this PR does not authorize activation.

@dcadenas the head moved after your approval, so I have re-requested your review.

NotThatKindOfDrLiz (approved):
Approved on `0f8eb69`. The remaining SHA extraction asymmetry is fixed in the authoritative VCL and runbook, the regression is proven red-green, all 84 script tests pass, and CI is green. This remains a source-only change; Fastly activation is still gated on divinevideo/divine-funnelcake#1070.

PR #190, opened 2026-08-10
fix(transcription): reject prompt leakage in captions

### LINKED ISSUE #189: Prevent transcription prompt leakage in published captions
## Problem\n\nSome generated WebVTT captions contain verbatim transcription prompt or schema instructions instead of spoken audio. Music-only clips can also receive hallucinated prose instead of a closed-caption sound cue.\n\n## Expected behavior\n\n- Prompt/schema commentary is rejected before publication.\n- Confident music and other non-speech classifications use standard bracketed closed-caption cues.\n- Silence or uncertain classifications publish no invented caption.\n- Existing leaked prompt text can be found by the repair scanner.\n\n## Validation\n\nAdd regression coverage for the observed production phrase, constrained sound-event handling, silence, and unknown labels.

rabble:
## Summary

- remove the leaked explanatory paragraph from the Gemini transcription prompt
- reject echoes of the prompt we currently send before publishing WebVTT, and teach the repair scanner to find existing instances of the retired prompt text
- classify no-speech Gemini responses into constrained closed-caption cues such as `[Music]`, while leaving silence and unknown labels empty

## Motivation

A production caption displayed verbatim prompt text instead of audio transcription. Music-only clips must not expose model instructions or invented prose.

## Linked issue

Closes #189

## Manual validation

- `cargo test --manifest-path cloud-run-transcoder/Cargo.toml --locked` — 153 passed
- `python3 -m unittest tests.test_scan_and_repair_vtts tests.test_repair_recent_bad_vtts` (from `cloud-run-transcoder/`) — 25 passed
- `python3 -m unittest discover -s tests -p 'test_*.py'` (from `cloud-run-transcoder/`, the command CI declares) — 30 passed
- `cargo clippy --manifest-path cloud-run-transcoder/Cargo.toml --locked --all-targets --all-features` — passed with existing warnings
- `git diff --check` — passed

No visual layout change. The visible behavior change is that confidently classified non-speech audio uses standard bracketed CC cues and rejected model leakage is not published.

## Takeover review follow-up (dw-review-start leaf)

A requested-review takeover leaf added four commits on top of `b023b3f`, all within `cloud-run-transcoder/`:

- `56cce86` — closes a live-gate miss: a single-bullet echo of the prefix/suffix instruction was published verbatim; added an exact prompt-echo marker for it (both languages' lists, parity extractor hardened).
- `83f834e` — corrects a comment in the echo safety test that contradicted the retired-marker split.
- `0cc064e` — replaced the new marker with a longer self-identifying span after second-judgment re-review showed the shorter one could flag real speech quoting LLM formatting advice; added boundary speech samples to both safety tests.
- `8945727` — removes an unterminated rustdoc code fence the docstring update introduced.

Second-judgment review ran four rounds on the final head `8945727e` and converged on ready with no open findings. Validation at head: transcoder Rust tests 165 pass, transcoder Python suites 30 + 93 pass, clippy warnings unchanged (pre-existing only), `cargo doc` codeblock warnings clean.


cloud-run-transcoder/src/main.rs
```diff
@@ -3742,6 +3776,25 @@ fn distinct_marker_phrases(normalized: &str, markers: &[&str]) -> usize {
...
+    // These are verbatim fragments of prompts previously sent by this
+    // service. Unlike generic schema language, a single match is conclusive:
+    // this prose is not audio transcription. Keep them so the repair scanner
+    // can identify VTTs produced before the prompt was shortened.
+    const EXACT_PROMPT_MARKERS: &[&str] = &[
+        "this transcript is consumed by an automated caption pipeline",
+        "can only parse the exact json shape below",
+        "the pipeline cannot recover the captions",
+        "strict adherence to the format below is what makes that possible",
+        "classify the dominant sound in sound_event",
```
NotThatKindOfDrLiz: This marker does not match the prompt it is meant to catch. The live prompt says `classify the dominant sound in `sound_event`` with backticks around `sound_event`, and `normalize_for_marker_scan` preserves punctuation, so `normalized.contains("classify the dominant sound in sound_event")` is false. Please either include the backticks in the marker or shorten it to a punctuation-independent fragment like `classify the dominant sound in`. The Python scanner copy needs the same change.

cloud-run-transcoder/scan_and_repair_vtts.py
```diff
@@ -68,6 +68,18 @@
...
 
+# Verbatim fragments of prompts previously sent by the transcoder. One match is
+# conclusive and intentionally catches already-published VTTs after the prompt
+# itself has been shortened.
+INSTRUCTION_ECHO_EXACT_PROMPT_MARKERS = (
+    "this transcript is consumed by an automated caption pipeline",
+    "can only parse the exact json shape below",
+    "the pipeline cannot recover the captions",
+    "strict adherence to the format below is what makes that possible",
+    "classify the dominant sound in sound_event",
```
NotThatKindOfDrLiz: Same backtick mismatch as the Rust guard: the prompt contains `` `sound_event` ``, but this exact marker omits the backticks and the scanner normalizer does not strip punctuation. Please keep this exact-marker list byte-identical with Rust and add parity coverage for it, like the existing `STRONG_MARKERS` parity test.

cloud-run-transcoder/src/main.rs
```diff
@@ -3489,7 +3487,30 @@ fn normalize_transcript_to_vtt(raw: &str) -> Result<ParsedVtt> {
         }
     }
 
-    // If we parsed valid JSON but found no segments and no text,
+    // Convert a constrained non-speech classification into conventional
+    // closed-caption notation. Only do this when no speech cue was produced;
+    // a generic sound label must never replace intelligible speech.
+    if let Ok(json) = serde_json::from_str::<serde_json::Value>(trimmed) {
+        let segments_empty = json["segments"]
```
NotThatKindOfDrLiz: This gates on raw JSON array emptiness, but the comment says the sound cue should be used when no speech cue was produced. If Gemini returns `segments` with only empty/whitespace text plus `sound_event: "music"`, the segment loop emits no cue, then this block skips `[Music]` because the array is non-empty. Please move this into the existing parsed-JSON block and gate on `cue_index == 1`/no emitted cue instead. That also avoids parsing the same JSON a third time.

NotThatKindOfDrLiz (changes requested):
Thanks for removing the leaked motivational paragraph and for constraining non-speech captions through an allowlist. I think this needs a small revision before merge: one exact prompt marker is currently inert, the new Rust/Python exact-marker list lacks parity/live-prompt coverage, and the sound-event fallback should be gated on whether no cue was emitted rather than whether the raw segments array is empty.

cloud-run-transcoder/src/main.rs
```diff
@@ -3721,6 +3757,84 @@ fn distinct_marker_phrases(normalized: &str, markers: &[&str]) -> usize {
...
+/// published — the incident this list exists for.
+///
+/// If the old prompt is ever restored, `retired_prompt_markers_are_no_longer_sent`
+/// fails and forces the affected marker into `EXACT_PROMPT_MARKERS_CURRENT`, where
+/// it would be live-gated and bound by the invariants above. That test, not this
+/// list, is the live safety net.
+///
+/// Keep byte-identical with `INSTRUCTION_ECHO_EXACT_PROMPT_MARKERS_RETIRED` in
+/// `scan_and_repair_vtts.py` (CI asserts parity).
+const EXACT_PROMPT_MARKERS_RETIRED: &[&str] = &[
```
dcadenas, line 3933: Retired prompt prose is deliberately not scanned here, and that is the one change worth pausing on.

`build_gemini_prompt` is this service's only prompt builder and no longer emits this paragraph, so a match in the live gate could never be a real echo — only a real speaker discussing caption pipelines. Because a single exact-marker match is conclusive and the resulting drop is terminal (empty VTT, `status=complete`, edge-cached, no auto-retranscribe), keeping these in the publish gate bought zero detection for a permanent-caption-loss risk.

These phrases are prose about captions and JSON, so lengthening them cannot close the class — any such phrase is utterable by someone talking about captions and JSON. Scope closes it.

The scanner still carries them, which is where they can actually fire, and `retired_prompt_markers_are_no_longer_sent` is the safety net: restore the old prompt and it fails, forcing the marker into the live list.

cloud-run-transcoder/src/main.rs
```diff
@@ -3487,9 +3485,34 @@ fn normalize_transcript_to_vtt(raw: &str) -> Result<ParsedVtt> {
...
             }
         }
+
+        // Reaching here means no speech cue was produced. Convert a
+        // constrained non-speech classification into conventional
+        // closed-caption notation. Gated on "no cue emitted" rather than "the
+        // segments array is empty" so a response carrying only blank segments
+        // still yields the sound cue; a generic sound label must never replace
+        // intelligible speech.
+        if let Some(label) = closed_caption_sound_label(json["sound_event"].as_str()) {
```
dcadenas, line 3592: Keeping `confidence: None` here rather than forwarding the parsed confidence is deliberate.

`extract_transcript_confidence` collects any `logprob` values in the response, and `transcript_drop_reason` drops a transcript whose confidence reads low. For audio with no speech, low speech confidence is the expected reading — so forwarding it would let a correct `[Music]` cue be dropped as `LowProviderConfidence`, terminally.

This is latent rather than observed: the Vertex request never sets `responseLogprobs`, so confidence is `None` in production today. `sound_cue_does_not_carry_speech_confidence` pins it against a future change that starts requesting logprobs.

dcadenas (approved):
✅ The prompt-leak fix and the constrained CC cues are sound — I pushed fixes for the three review findings plus two defects found along the way.

The allowlist approach to non-speech cues is the right shape: `silence`, `none`, and unknown labels producing no cue is what keeps a bad classification from inventing captions.

### Changes I pushed

- Gated the sound cue on whether a cue was emitted rather than on the raw `segments` array being empty. A response with blank-text segments plus `sound_event: "music"` previously lost its `[Music]` cue and errored as "no usable text".
- Made the `sound_event` marker byte-identical to the live prompt instead of truncating it past the backticks.
- Stopped the sound cue from carrying the provider's speech confidence.
- Extended the retired prompt markers past phrases an ordinary speaker could say, and added parity plus live-prompt coverage so a marker cannot silently go inert again.
- Broadened retired-marker coverage to every sentence of the removed paragraph, pinned as a fixture since the live prompt no longer carries the text.

Retired prompt prose no longer runs in the live publish gate. It is the widest change here, it is isolated in its own commit, and the reasoning is on the marker list inline. The cost is real: the scanner now flags more than the gate drops, so a clip whose genuine speech contains one of those spans will be re-transcribed on each scanner run and come back unchanged.

### Checks

- `cargo check --tests --locked`, `cargo clippy --locked --all-targets --all-features`
- `cargo test -p blossom-core --locked`, upload service tests
- `cargo test --manifest-path cloud-run-transcoder/Cargo.toml --locked`: 153 passed
- `git diff --check`: clean

CI declares two `python3 -m unittest discover` steps that no local check runner executes, so I ran them by hand: 30 tests in `cloud-run-transcoder/tests`, 71 in `scripts/tests`, both passing. Those are hand-run results, not covered by the declared check set.

`cargo fmt --check` reports six differences. All predate this branch or are the untouched-test formatting I restored, and `cargo fmt` is not a declared check here.

### Record correction

The description listed 143 Rust tests and 19 Python tests, and described the guard as rejecting exact prompt echoes generally. I updated both: the counts are current, and echoes of the *current* prompt are what the publish gate rejects.

NotThatKindOfDrLiz (approved):
The requested changes are addressed. I pushed the remaining mechanical cleanup in 2005e47, verified locally, and CI is green on the new head. Approving; rabble should review the takeover commits and merge when ready.

NotThatKindOfDrLiz:
@rabble I pushed one mechanical takeover fix in 2005e47: marked the retired prompt marker list as test-only so non-test transcoder builds no longer gain a new dead-code warning. The marker list is still available to the Rust tests, and the Python scanner parity tests still read it from source text.

Local verification:
- `cargo check --manifest-path cloud-run-transcoder/Cargo.toml --locked` — passed with only the pre-existing `VideoInfo.width/height` warning
- `cargo test --manifest-path cloud-run-transcoder/Cargo.toml --locked` — 153 passed
- `python3 -m unittest discover -s tests -p 'test_*.py'` from `cloud-run-transcoder/` — 30 passed
- `git diff --check` — clean

GitHub checks are green on the pushed head. Please review the takeover commit and merge when you are ready.

rabble:
Updated the branch to current main in merge commit 3600a9a without rewriting history. The only conflict was the cloud-run-transcoder test import list; the resolution retains both this PR’s prompt-leakage helpers/markers and main’s newer token/provider helpers.\n\nFresh validation on the merged result:\n- cargo fmt --manifest-path cloud-run-transcoder/Cargo.toml -- --check\n- cargo check --manifest-path cloud-run-transcoder/Cargo.toml --locked\n- cargo test --manifest-path cloud-run-transcoder/Cargo.toml --locked (165 passed)\n- python3 -m unittest discover -s tests -p test_*.py (30 passed)\n- scoped diff checks passed\n\nThe prior approval predates this merge commit, so I am requesting fresh platform review on the updated head.

rabble:
Integrated current `origin/main` as merge commit `b023b3f`, preserving the branch history. The VCL conflict resolved to the current main policy; the resulting tree differs from main only in this PR's three transcoder files.

Fresh merged-result validation passed: `cargo check --tests --locked`, 186 blossom-core tests, 41 upload tests, 165 transcoder tests, both Python suites, clippy, and `git diff --check`. Local Viceroy was unavailable; the required CI edge-test job is running on this exact head.

@divinevideo/platform: please review the current head; earlier approvals are now stale.

rabble:
@NotThatKindOfDrLiz please review and approve exact current head 8945727. I re-reviewed the four commits added after the prior validated head: the Rust/Python marker parity, exact prompt marker, speech-neighbour false-positive coverage, and doc correction are coherent. Targeted Rust instruction-echo tests (10), all transcoder Python tests (30), and diff check pass; CI is running on this head.

dcadenas (approved):
✅ The leakage gate is sound, and the marker coverage now matches what the docstring claims.

I reviewed the branch and pushed four fixes on top of `b023b3f`, all inside `cloud-run-transcoder/`.

### Changes I pushed

- `56cce86`: the live echo gate covered only two of the prompt's three instruction lines. A single-bullet echo of the prefix/suffix instruction passed the gate and would be published verbatim as captions (confirmed against the live predicate). Added an exact marker covering it to both marker lists, and hardened the test-side extractor so quoted markers compare as decoded text rather than Rust source.
- `83f834e`: corrected a comment in the echo safety test that claimed retired markers run in the live gate. They do not, and the test twelve lines below proves it.
- `0cc064e`: replaced the marker added in `56cce86` with a longer span of the same bullet. The first choice, `no "here is the json" preamble`, is generic LLM-formatting speech: a speaker quoting that phrase would lose captions for the whole clip (verified: `There should be no "Here is the JSON" preamble in the reply, right?` flagged). The replacement span is quote-free and self-identifying, and both safety tests now pin that boundary with speech samples.
- `8945727`: removed an unterminated rustdoc code fence the docstring update introduced.

### What was declined, and why

- Sound-cue webhook payload (`duration_ms: 0`, provider-detected language on a `[Music]` cue): the payload difference only exists where the PR publishes a cue at all, so nothing previously published changes. Zero duration already has in-repo precedent on the plain-text fallback path, and the consumer lives outside this repository.
- `enum` in the Vertex `responseSchema`: a documented, officially sampled feature for `application/json` responses, and a failure would be a loud 400 on every transcription request rather than silent corruption.
- A lone punctuation-stripped echo of the prompt's trailing sentences still passes the gate. That residual is documented in the marker list's invariants and accepted deliberately: closing it requires a fragment short enough to flag real speech, and this gate's rule is that real speech never loses captions.

### Checks

- `cargo test --manifest-path cloud-run-transcoder/Cargo.toml --locked`: 165 passed
- Transcoder Python suites: 30 and 93 passed
- `cargo clippy --locked --all-targets --all-features`: 12 pre-existing warnings, unchanged
- `cargo doc --no-deps`: no codeblock-attribute warnings
- CI at review time: `license/cla` and `semantic-pull-request` pass; `test` pending on the new head (run 32101369312)
- Edge crate tests not run: `viceroy` is not installed here, and this PR does not touch the edge.

### Review receipts

- Reviewed head `8945727e`, matching the live PR head at review time; live head before the push was `b023b3f`.
- Second review converged on ready after four rounds on this head lineage, with no open findings; two declines, listed above.
- Validation coverage: transcoder crate and both Python suites. Gaps: the edge crate suite, noted in Checks.
- Coordinator: `divine-work-coord` ask `01a0131e-103e-79c0-9b28-de8028d5473b`. Run id `sj-round-8945727ee48e3dab78d9cbc40407866258676b19`.

### DANIEL HANDOFF

This PR gates transcription responses that leak their own prompt into captions: a marker gate on the live path, an offline `scan_and_repair_vtts.py` that scans stored VTTs and repairs or quarantines, and parity tests keeping the Rust and Python marker lists identical.

On top of the original work I fixed one real miss (a single-bullet echo of the prefix/suffix instruction was publishable), replaced my first marker choice after re-review showed it could flag a real speaker quoting LLM advice, corrected one wrong comment, and fixed one rustdoc warning. Second review converged on ready with no open findings at `8945727e`.

Known accepted residual: a lone echo of the prompt's trailing sentences with its punctuation stripped still passes the gate, by design (recall is sacrificed before real speech is dropped). Everything else raised across the rounds is either fixed or declined with reasons above.

Merge-ready from this leaf's perspective once the pending `test` run on `8945727e` completes.

NotThatKindOfDrLiz:
## Re-review of head 8945727

All three prior findings are resolved and tests pass on `8945727`. The marker parity, live-prompt coverage, and sound-cue gating are all correct.

### One substantive issue: marker breadth

Commit `0cc064e` adds to the **live, single-match-conclusive** gate:

```
"do not prefix or suffix the json with any text, markdown, or explanation"
```

This closes a real gap — I confirmed a lone echo of that bullet escapes `STRONG_MARKERS` and would publish verbatim. But this marker violates **invariant 2** that the docstring documents (line 3882):

> Each marker must be long enough to self-identify as this pipeline's prompt.

"do not prefix or suffix" is generic JSON boilerplate in thousands of public prompts. A clip where someone reads a prompt aloud loses **all captions permanently** (`status=complete`, edge-cached, no auto-retranscribe).

Commit `0cc064e` rejected `no "here is the json" preamble` for this exact reason, then replaced it with a longer span that has the same property. The test samples pin paraphrases, not the verbatim bullet.

**Recommended fix:** extend the marker to the full bullet:

\`\`\`rust
"do not prefix or suffix the json with any text, markdown, or explanation. no \"here is the json\" preamble. no \`\`\` fences."
\`\`\`

This stays byte-identical to the prompt, remains detectable (a model echoing that bullet echoes the whole thing), and the false-positive surface drops to near zero. Both parity tests cover it already. Python gets the same one-line edit.

**Alternatively:** approve as-is with this recorded. The marker is measurably closer to safe than before.

Either way: CI is green, approvals are current, the sentinel fix is solid.

My recommendation: apply the marker widening, then approve.

PR #191, opened 2026-08-11
feat(storage): serve blobs from Fastly Object Storage with GCS fallback

rabble:
## Summary

Serves media blobs from Fastly Object Storage (FOS) when present, falls back to GCS when not, and lazily copies GCS-served objects into FOS so subsequent reads come from the replica. Both behaviours are behind independent config flags that default OFF.

GCS remains authoritative. This changes **where bytes are read from**, nothing else — moderation, access control, `Surrogate-Key`/`Surrogate-Control`, and range handling are untouched.

## Motivation

Origin egress is the dominant media cost at scale. Measured on our own account: **$0.182/GB blended** (APAC $0.176, AU $0.310, and APAC is 27% of bytes), ~4.6 TB/month, ~$838/month today. Extrapolated to launch-scale DAU that becomes the largest single line item. FOS has zero egress to the Fastly CDN, which removes it entirely.

Cache tuning cannot reach this: byte offload is capped by content reuse. We raised it from 25.8% to 37.2% with range caching earlier, and the remainder is long-tail.

## Verified on staging, not just in review

Deployed to `fastly-blossom-staging` (`xsfh87UGXl8NUmgsbpotYj`) with both flags **on**, pointed at production buckets read-only:

- An object **verifiably absent** from FOS was requested → served correctly (200, 732,457 bytes, `x-sha256` matching) → **appeared in FOS at the identical byte count**.
- That single test exercises the whole chain: FOS read attempted → miss → GCS fallback → write-back. It also confirms SigV4 signing, region, path-style addressing, backend and credentials, none of which could be verified locally (the crate is wasm-only).
- Objects already mirrored serve correctly with matching `x-sha256`.

100,000 objects (~80.7% of all views) are already mirrored into `divine-media-delivery` and sha256-verified against their keys.

## Testing

- `cargo test -p blossom-core --locked` — 131 passed, 0 failed (16 new)
- `cargo check --tests --locked` — clean, no new warnings
- `cargo clippy --locked --all-targets --all-features` — no warnings from touched code

Pure decision logic lives in `blossom-core/src/read_through.rs` **because tests in the edge crate compile but never execute** (CI runs `cargo check --tests`; the crate cannot link natively). Flag parsing, write-back eligibility, size bounds and path construction are all tested there where tests actually run.

## Risk and rollout

- Both flags default OFF. Absent or unparseable config means disabled, never enabled.
- Any FOS error, timeout or non-200 falls through to GCS silently. FOS cannot break media delivery.
- Write-back is bounded at 32 MB (largest observed object: 27.9 MB) and skips silently above it.

Suggested rollout: enable `fos_read_enabled` alone first and watch `[FOS] miss` logs; leave `fos_write_back_enabled` off until the item below is resolved.

## Known gaps, deliberately not fixed here

1. **Deletes do not propagate to FOS.** Takedown removes from GCS only, so FOS copies orphan. This gap already exists from the out-of-band mirror; write-back grows it. Serving stays safe because metadata access checks run before any storage read. A FOS delete should land before write-back is enabled in production.
2. **Write-back buffers** an otherwise-streamed response (peak ~2× object size, well inside the 128 MB limit), raising p99 cache-fill latency. This partially works against the streaming-miss change made separately.
3. Audio read paths were deliberately left on the existing GCS-only function.

## Manual validation plan

1. Merge with both flags off — no behaviour change.
2. Set `fos_read_enabled=true`, confirm media serves and `[FOS] miss` appears for unmirrored hashes.
3. Confirm byte offload and error rates unchanged.
4. Only after a FOS delete path exists, enable `fos_write_back_enabled`.

No visual change — this is a storage-layer change with identical response bodies and headers.

Follow-up: #192

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -62,3 +62,7 @@ scripts/backfill-orientation-failed.txt
 .claude/
 __pycache__/
 *.log
+
+# Local storage credentials (never commit)
+.fastly-fos-objects-key.json
+.gcs-hmac-mirror.json
--- a/blossom-core/src/lib.rs
+++ b/blossom-core/src/lib.rs
@@ -1,5 +1,6 @@
 pub mod delete_policy;
 pub mod error;
 pub mod rate_limit;
+pub mod read_through;
 pub mod transcribe;
 pub mod types;
--- a/blossom-core/src/read_through.rs
+++ b/blossom-core/src/read_through.rs
@@ -1,0 +1,299 @@
+// ABOUTME: Pure decision logic for read-through delivery from a mirror origin.
+// ABOUTME: Feature-flag parsing, write-back eligibility, and shared object paths.
+
+/// Source label for objects served from the authoritative GCS bucket.
+pub const SOURCE_GCS: &str = "gcs";
+
+/// Source label for objects served from the Fastly Object Storage mirror.
+pub const SOURCE_FOS: &str = "fos";
+
+/// Largest object we are willing to buffer in memory for a write-back.
+///
+/// Fastly Compute allows 128 MB of memory per request and a write-back holds
+/// the object twice for a moment (once for the upload body, once for the
+/// response body), so the ceiling has to leave generous headroom. The largest
+/// media object observed in the catalogue is 27.9 MB.
+pub const MAX_WRITE_BACK_BYTES: u64 = 32 * 1024 * 1024;
+
+/// Parse a config-store feature flag.
+///
+/// A flag is enabled only when it is present and explicitly affirmative.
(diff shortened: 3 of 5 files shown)


rabble:
Follow-up filed as #192 — deletes and takedowns do not propagate to FOS.

**Do not enable `fos_write_back_enabled` in production until #192 items 1–3 land.** `fos_read_enabled` is safe to enable independently; reading from a replica creates no new orphans.

Note the orphan gap already exists from the out-of-band bulk mirror (100,000 objects), so #192 is worth doing regardless of whether write-back is ever turned on.

PR #193, opened 2026-08-11
docs(delivery): select B2 for second CDN origin

rabble:
## Summary

- select a private Backblaze B2 bucket as the approved-only origin for the second CDN
- keep GCS authoritative and Fastly Object Storage scoped to the Fastly delivery path
- require permanent deletion of every B2 file version plus cache purges for takedowns
- record why B2 was chosen over R2, and mark the partner-CDN egress rate unverified with its downside case
- steer a hash to the second CDN only after its replica copy verifies, so freshly approved content is not steered before it exists
- name DNS on a Divine-owned delivery hostname as the retroactive rollback lever; setting rollout to zero stops new URLs but does not retire issued ones, and it is not automatic request failover

The adjacent docs are updated in the same change so the repo gives one answer: the steering design's selector, eligibility rule, origin table, and rollback section; the vendor notes' open question, narrowed to bunny's AWS-signing path to a private bucket; the status entry point, which now indexes this decision; and the ACL validation record, which no longer reads as ruling out a private origin.

The pilot must re-prove the browser-visible behavior — CORS for range and subtitle reads, `Content-Type` carry-through, cache-control alignment — because the validated probe bucket was public.

## Motivation

Resolve the origin decision that blocked the multi-provider pilot while preserving the moderation boundary and the deletion behavior proven by the existing takedown drill.

## Linked issue

Related to #177. Closes no issue.

## Manual validation

- `git diff --check origin/main...HEAD`
- compared the plan against the existing B2 takedown drill, the vendor capability notes, and the bunny rollout plan on main
- `cargo check --tests --locked`, `cargo test -p blossom-core --locked`, the upload and transcoder suites, and `cargo clippy --locked --all-targets --all-features`
- the declared Python suites: 23 tests in `cloud-run-transcoder`, 71 in `scripts/tests`
- documentation only; no runtime or visual change

### DIFF
--- a/docs/superpowers/plans/2026-08-07-blossom-multi-provider-delivery-policy.md
+++ b/docs/superpowers/plans/2026-08-07-blossom-multi-provider-delivery-policy.md
@@ -104,7 +104,14 @@ replicated, and later banned or deleted.
 
 After a blob is approved and public, explicitly copy it to B2 from the existing pipeline. Verify
 length and SHA-256. Record which providers hold it as one field on the existing metadata record.
-Do not block publication on the copy.
+
+Publication does not wait for the copy, but **steering does**. Set the replicated flag only after
+the copy verifies, and route a hash to bunny only once that flag is set. Reversing this order 404s
+freshly-approved content for its first viewers. The implementation contract for that ordering — the
+queued job, the flag, and the reconciler that asserts replica presence for `Active` hashes and
+absence for everything else — is the bunny rollout plan
+(`2026-08-07-bunny-delivery-rollout-plan.md`); this document selects the origin and does not restate
+or override it.
 
 Do not configure a pull-through or automatic rehydration path from preserved GCS objects into B2. A
 deleted replica must not be silently recreated after a moderation action. The writer gets a
@@ -112,8 +119,9 @@ bucket-scoped credential with the explicit upload and version-delete capabilitie
 gets a separate bucket-scoped read-only credential. Neither credential belongs in source control or
 client-visible configuration.
 
-No queue, no reconciler, no replication framework, no BUD-04 `PUT /mirror` path. A periodic repair
-command covers gaps until measurement shows it does not.
+No generic replication framework and no BUD-04 `PUT /mirror` path. The queue is the existing
+derivative-status queue and the reconciler is a periodic repair pass over one flag — reuse, not new
+machinery.
 
 Recheck policy immediately before copying. A delayed job must never publish a hash that has since
 been tombstoned or restricted.
--- a/docs/superpowers/plans/2026-08-07-blossom-multi-provider-delivery-policy.md
+++ b/docs/superpowers/plans/2026-08-07-blossom-multi-provider-delivery-policy.md
@@ -4,7 +4,7 @@
 # Blossom Multi-Provider Delivery Plan
 
 **Date:** 2026-08-07 (rewritten to cut scope)
-**Status:** Draft — one decision blocks the rest
+**Status:** Decision recorded — use Backblaze B2 as the second-CDN origin; pilot remains unapproved
 **Scope:** Public read path only. Not authorization to deploy.
 **Related:** `../specs/2026-08-06-cdn-delivery-steering-design.md`,
 `../specs/2026-08-06-media-delivery-cost-architecture-review.md`
@@ -34,27 +34,43 @@ the whole pilot. Everything else in the earlier draft is deferred at the bottom 
   Stream is out of scope. If a proposal starts moving processing back to a CDN vendor, it has
   stopped being this plan.
 
-## The decision that blocks everything else
+## Origin decision — private Backblaze B2
 
-The second CDN needs an origin it can read at delivery rate. Fastly Object Storage's *public* S3
-endpoint is documented at "~150Mbps and limited to a total of 100 requests per second"
-([Fastly product docs](https://docs.fastly.com/products/object-storage)) — about 1.6 TB/day, against
-a ~37 TB/day origin read requirement at 1M DAU. Reads from inside Fastly's network (Compute,
-Deliver) are exempt, so this does not affect the sibling FOS-behind-Fastly plan. It only bites when
-a *non-Fastly* CDN reads the bucket.
+**Use a private Backblaze B2 bucket as the approved-only replica origin for bunny.** GCS remains
+Divine's private authoritative store. Fastly Object Storage remains a separate, Fastly-only delivery
+mirror. Do not put a non-Fastly CDN in front of FOS unless Fastly quotes and contractually supports
+the required public-endpoint capacity.
 
-The same docs say "If your needs exceed these limits, reach out to sales@fastly.com," so this is a
-commercial question, not a hard ceiling — but **a raise nobody has quoted is not a capacity plan.**
-Until Fastly gives a sustained number in writing (cost review open question 16), assume FOS cannot
-be the second CDN's origin. Two workable shapes:
+Why B2 instead of letting bunny read GCS directly:
 
-- **(a)** second CDN reads GCS directly — works today, but GCS egress returns on every cache miss;
-- **(b)** R2 as the replica origin, $0 egress to any consumer — what the cost review recommends
-  first (§8.1 step 5).
+- Backblaze lists bunny as a partner CDN with unlimited free egress and prices B2 at $6.95/TB-month
+  with free Class A/B/C transactions
+  ([B2 pricing](https://www.backblaze.com/cloud-storage/pricing)). GCS charges network egress on
+  every bunny cache miss, which preserves the cost line this pilot exists to remove.
+- B2 has a private S3-compatible endpoint and bucket-scoped application keys. bunny pull zones
+  expose AWS signing configuration
+  (`AWSSigningEnabled`, key, secret, and region) in the
+  [pull-zone API](https://docs.bunny.net/api-reference/core/pull-zone/add-pull-zone), so the pilot
+  can keep the replica private. Prove the B2 region, endpoint, host/path style, and range behavior
+  against a real pull zone before traffic.
+- B2 is always versioned. An S3 `DeleteObject` or `b2_hide_file` can leave older bytes retrievable;
+  takedown must enumerate and permanently delete every file version
+  ([B2 file versions](https://www.backblaze.com/docs/cloud-storage-file-versions)).
+- B2 is a disposable replica. Its availability does not replace GCS durability because rollout can
+  be set to zero and Fastly/GCS remains the authoritative path.
 
-Pick one before building anything. The earlier "Bunny primary + Fastly Object Storage standby"
-shape is not buildable as written: it names FOS as the standby origin *and* assumes a second CDN in
-front of it.
+The initial GCS-to-B2 copy incurs one-time GCS egress; measure that separately from steady-state
+delivery cost. Confirm on the production billing account that bunny-origin reads receive the
+advertised partner-CDN egress treatment before ramping traffic.
+
+GCS remains acceptable only as a short-lived non-production diagnostic origin if needed to test
+bunny's AWS-signing compatibility. It is not the production choice. The earlier "Bunny primary +
+Fastly Object Storage standby" shape conflated two independent mirrors; the settled topology is:
+
+```text
+GCS private authority -> FOS private mirror -> Fastly
+                      +-> B2 approved-only mirror -> bunny
+```
 
 ## The scope rule that removes most of the work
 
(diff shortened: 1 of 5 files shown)


rabble (commented):
Verified against the sources it depends on.

- B2 claims ($6.95/TB-mo, free Class A/B/C transactions, partner-CDN egress) match the repo's verified vendor notes (`docs/cdn-object-storage-vendor-notes.md`, marked [V]), and the doc honestly carries the [U] caveat — billing-account confirmation of partner egress is in the go/no-go list.
- The version-deletion takedown ordering matches the proven drill (`docs/measurements/2026-08-07-takedown-drill.md`: hide does not erase).
- The private cost review it cites exists where the status doc says it lives (divine-context `repo-context/` on main) and this decision is consistent with it, including F2.
- Rollback semantics are consistent with #184: percentage-to-zero is roll-forward control for new descriptors; the DNS move on the owned hostname is the retroactive one. Both docs say so without contradicting each other.
- No-secret check: names the private repo path only, no commercial figures. Clean.

Pre-existing, not introduced here: the `Related:` link to `../specs/2026-08-06-media-delivery-cost-architecture-review.md` targets a path that can never exist in this public repo (commercial analysis is deliberately kept in divine-context). Worth its own small fix rather than expanding this PR.

No findings. The 'roll-forward control, not request-level failover' correction is the important honesty fix in this revision.

docs/superpowers/plans/2026-08-07-blossom-multi-provider-delivery-policy.md
```diff
@@ -34,27 +34,43 @@ the whole pilot. Everything else in the earlier draft is deferred at the bottom
...
-front of it.
+## Origin decision — private Backblaze B2
+
+**Use a private Backblaze B2 bucket as the approved-only replica origin for bunny.** GCS remains
+Divine's private authoritative store. Fastly Object Storage remains a separate, Fastly-only delivery
+mirror. Do not put a non-Fastly CDN in front of FOS unless Fastly quotes and contractually supports
+the required public-endpoint capacity.
+
+Why B2 instead of letting bunny read GCS directly:
+
```
NotThatKindOfDrLiz, line 46: This now explains why B2 beats GCS, but it drops the prior R2 recommendation without rebutting it. The existing public notes say R2 has verified $0 egress to any consumer plus published durability/SLA, while B2 partner-CDN egress still needs production billing confirmation. If B2 is the chosen origin, please add the short B2-over-R2 rationale here so the decision trail survives this rewrite.

docs/superpowers/plans/2026-08-07-blossom-multi-provider-delivery-policy.md
```diff
@@ -87,44 +109,65 @@ been tombstoned or restricted.
 ## Steering
 
 Server-side only, per the steering design: a `select_delivery_host()` call at the descriptor seam.
-Ineligible content never gets the second host. The percentage lives in the config store; rollback is
-setting it to 0. No mobile or web client change.
+Ineligible content never gets the second host. The percentage lives in the config store. Setting it
+to 0 stops issuing new bunny URLs without a deploy.
+
+That is **roll-forward control, not request-level failover**. A client that already holds a bunny
+URL continues using it until it refreshes the descriptor. The pilot must measure and bound that URL
```
NotThatKindOfDrLiz: This rollback model conflicts with the merged rollout plan's owned-hostname requirement. That plan says not to emit a vendor hostname; descriptors should use an owned host such as `v.divine.video`, with CDN rollback handled by DNS at a short TTL. Please rewrite this around the two separate levers: setting rollout to 0 stops newly eligible descriptors, while a DNS flip handles already-issued owned delivery URLs.

docs/superpowers/plans/2026-08-07-blossom-multi-provider-delivery-policy.md
```diff
@@ -87,44 +109,65 @@ been tombstoned or restricted.
...
+on partial completion. If the required legal SLA is shorter than this fan-out can guarantee, add an
+edge deny-list before increasing rollout; descriptor steering alone is insufficient.
 
 ## Pilot
 
 20–100 representative approved public objects.
 
-1. Non-production bucket/zone and credentials at the second provider.
+1. Private non-production B2 bucket, separate scoped writer and read-only origin
+   credentials, and a bunny pull zone with AWS signing enabled.
```
NotThatKindOfDrLiz: Please include the browser-visible B2 requirements from the rollout plan in the prove-first list. The validated public-bucket setup had explicit CORS behavior for range/VTT reads and upload-time cache-control alignment; moving to a private S3-signed origin should prove those still hold, otherwise video can play while subtitle tracks silently fail or cache semantics diverge.

docs/superpowers/plans/2026-08-07-blossom-multi-provider-delivery-policy.md
```diff
@@ -146,7 +189,8 @@ Each of these needs its own justification. None belong in the pilot.
 
 ## Open questions
```
NotThatKindOfDrLiz, line 231: Since this PR records the origin decision, please update the adjacent docs that still present the decision as open: the status entry point should index this policy and record B2/private-origin status, the steering design should stop listing R2 and B2 as co-equal unresolved options, and the vendor notes open question about private R2/B2 origins should reflect whatever this PR settles.

NotThatKindOfDrLiz (changes requested):
Requesting changes on the decision record before this merges. The takedown rewrite is a real improvement, but the origin decision doc needs to line up with the already-merged owned-hostname rollout model and preserve the reasoning trail for choosing B2 over the existing R2 recommendation.

Please also update the adjacent status/design/vendor-notes docs so the repo does not say the origin question is both settled and open.

docs/superpowers/plans/2026-08-07-blossom-multi-provider-delivery-policy.md
```diff
@@ -66,65 +94,111 @@ That single rule deletes, from the earlier draft:
...
-Ineligible content never gets the second host. The percentage lives in the config store; rollback is
-setting it to 0. No mobile or web client change.
+Ineligible content never gets the second host. The percentage lives in the config store. Setting it
+to 0 stops issuing new bunny URLs without a deploy.
+
+That is **roll-forward control, not request-level failover**. A client that already holds a
+second-provider URL continues using it until it refreshes the descriptor. The pilot must measure and
+bound that URL lifetime.
+
+The retroactive lever is DNS, and it only exists if the delivery URLs are on a Divine-owned
```
dcadenas, line 148: Rollback said already-issued URLs are retired by "descriptor refresh or explicit purge". A purge does not retire a URL — it forces bunny to refetch from B2, and during a rollback the object is still there, so bunny keeps serving. Deleting the replica instead would 404 the client, which is breakage rather than rollback.

The lever that works retroactively is DNS, and it only exists if descriptors carry a Divine-owned hostname. That matches the rollout plan on main, which stands up `v.divine.video` at stage 0a and forbids emitting a vendor hostname. Pilot step 1 now provisions that hostname, since step 5 exercises the move.

docs/superpowers/plans/2026-08-07-blossom-multi-provider-delivery-policy.md
```diff
@@ -66,65 +94,111 @@ That single rule deletes, from the earlier draft:
...
 
 ## Replication
 
-After a blob is approved and public, best-effort copy it to the second provider from the existing
-pipeline. Verify length and SHA-256. Record which providers hold it as one field on the existing
-metadata record. Do not block publication on the copy.
+After a blob is approved and public, explicitly copy it to B2 from the existing pipeline. Verify
+length and SHA-256. Record which providers hold it as one field on the existing metadata record.
+
+Publication does not wait for the copy, but **steering does**. Set the replicated flag only after
```
dcadenas, line 116: "Do not block publication on the copy" plus "No queue, no reconciler" contradicted the rollout plan on main, which enqueues the copy, sets the replicated flag only after verification, and routes to bunny only then — "the descriptor must not point at bunny until replication is confirmed, or a freshly-approved video 404s for its first viewers".

Split into the two conditions: publication does not wait, steering does. The queue, flag, and reconciler contract stays in the rollout plan rather than being restated here.

One detail worth keeping: the copy should follow the derivative-status queue's pattern rather than share that queue. It runs with `--max-concurrent-dispatches=1` as part of a generation-ordering contract, and replication has no generation.

docs/superpowers/plans/2026-08-07-blossom-multi-provider-delivery-policy.md
```diff
@@ -34,27 +35,54 @@ the whole pilot. Everything else in the earlier draft is deferred at the bottom
...
+  [pull-zone API](https://docs.bunny.net/api-reference/core/pull-zone/add-pull-zone), so the pilot
+  can keep the replica private. Prove the B2 region, endpoint, host/path style, and range behavior
+  against a real pull zone before traffic.
+- B2 is always versioned. An S3 `DeleteObject` or `b2_hide_file` can leave older bytes retrievable;
+  takedown must enumerate and permanently delete every file version
+  ([B2 file versions](https://www.backblaze.com/docs/cloud-storage-file-versions)).
+- B2 is a disposable replica. Its availability does not replace GCS durability because rollout can
+  be set to zero and Fastly/GCS remains the authoritative path.
+
+Why B2 and not R2, which the cost review ranks first (§8.1 step 5): on paper R2's $0 egress to any
```
dcadenas, line 74: The paragraph this replaced named R2 as what the cost review ranks first (§8.1 step 5), and the new section did not mention R2 at all. A decision record that drops the option its own cited source prefers loses the thing it exists to preserve.

The answer was already in the repo: B2 has been stood up and drilled end-to-end as a real pull-zone origin, while R2 has no bucket and no token. Recorded on those terms, with R2 named as the first fallback if partner-CDN egress does not appear on the billing account.

docs/superpowers/specs/2026-08-06-cdn-delivery-steering-design.md
```diff
@@ -52,6 +54,10 @@ fn select_delivery_host(req: &Request, meta: &BlobMetadata) -> &'static str {
     if !is_publicly_cacheable(meta) {
         return FASTLY_HOST;
     }
+    // Not yet copied and verified into the replica: bunny would 404 it.
+    if !meta.replicated {
```
dcadenas, line 59: The selector routed on eligibility alone, which reopens the 404 the plan's readiness rule closes: an approved blob that has not finished copying is absent from the replica for the same reason gated content is.

The ineligibility check still runs first, so a bug in the readiness leg cannot expose gated content. Both `is_publicly_cacheable` and `meta.replicated` are proposed symbols, and the note under the block now says so.

dcadenas (approved):
✅ The B2 selection holds up against the repo's own evidence — the takedown drill, the vendor notes, and the topology the rollout plan on main already assumes.

I pushed fixes for the places the document did not yet match that evidence.

### Changes I pushed

- Recorded why B2 was chosen over R2, the option the cited cost review ranks first.
- Marked the partner-CDN egress rate unverified and gave the downside case.
- Gated second-CDN steering on a verified replica copy, in both the plan and the steering design.
- Named DNS on a Divine-owned hostname as the retroactive rollback lever, and dropped cache purge as one.
- Pointed the cost-review reference at the copy that exists, and fixed the section number for the GA4 geography table.

The steering design is a merged spec rather than a file this branch set out to touch. Three of its claims disagreed with the decision recorded here — the rollback wording, the selector, and the goal line — so the corrections are isolated in their own commits.

### Checks

- `cargo check --tests --locked`, `cargo test -p blossom-core --locked`, upload and transcoder suites, `cargo clippy --locked --all-targets --all-features`
- declared Python suites: 23 tests in `cloud-run-transcoder`, 71 in `scripts/tests`
- CI green on the pushed head

### Follow-ups

- `docs/cdn-evaluation-status.md` is the declared entry point but indexes neither delivery plan. Pre-existing, and worth its own change rather than widening this one.

Merge decision is yours.

dcadenas (approved):
All four points from the change request are addressed on the branch.

The B2-over-R2 rationale and the owned-hostname rollback model landed earlier in the branch; the browser-visible requirements and the adjacent-doc sweep are new.

### Changes I pushed

- Prove-first list now covers CORS for range and subtitle reads, `Content-Type` carry-through, and cache-control set at upload, since the validated probe bucket was public and this selects a private signed origin.
- Steering design marks B2 selected and R2 the fallback instead of listing them as co-equal open options.
- Vendor notes narrow the open question to bunny's AWS-signing path to a private bucket, which is the part that is genuinely unproven.
- Status entry point indexes the decision and records it, on measured readiness rather than on cost.

### Correcting the record

The ACL validation record said a private bucket cannot back a pull zone. That is true of B2's native download path, where the authorizations are time-limited, and it is not true of the S3-compatible endpoint with AWS signing — the route selected here. As written it read as ruling out this origin. It now says unproven rather than excluded, which is what the evidence supports. The claim predates this branch.

Related: the rollout plan's CORS rule lists B2's native download operations, and a B2 CORS rule applies only to the operations it lists. Copying it could pass an origin-signing test while browser range and subtitle requests still fail. The pilot is now told to derive and record the S3-endpoint equivalent rather than reuse that rule.

### Checks

- `cargo check --tests --locked`, `cargo test -p blossom-core --locked`, upload and transcoder suites, `cargo clippy --locked --all-targets --all-features`
- declared Python suites: 23 tests in `cloud-run-transcoder`, 71 in `scripts/tests`
- CI green on the pushed head

My earlier approval was submitted against an older head before the change request was visible to me. This one is against the current head and does not clear that review — @NotThatKindOfDrLiz, over to you.

NotThatKindOfDrLiz:
@rabble I pushed one final docs-only takeover commit: `0ad461a docs(cdn): narrow private B2 origin-auth open question`.

What changed:
- narrowed the steering-design open question to bunny AWS signing against the private B2 S3 endpoint
- left R2 described only as the untested fallback, matching the origin table and vendor notes
- resolved the eight review threads after verifying the requested changes against the current head

Verification:
- `git diff --check origin/main...HEAD`
- CI green on the pushed head: `test`, `semantic-pull-request`, and CLA pass; deploy/docker jobs skipped as expected

Please review the takeover commit and merge when you are ready.

NotThatKindOfDrLiz (approved):
Approved after takeover.

I verified the current head against the requested changes and pushed the remaining one-line consistency fix in `0ad461a`: the steering design now narrows the open origin-auth question to private B2 AWS signing, with R2 marked only as untested fallback. That matches the plan, status doc, and vendor notes.

Checks are green on the pushed head. No further findings from my final pass.

PR #195, opened 2026-08-11
fix(transcoder): refresh expired provider tokens

### LINKED ISSUE #194: fix: Refresh expired metadata access tokens during transcription
## Summary
The Cloud Run transcoder can cache a metadata-server access token beyond the token remaining lifetime, causing transcription requests to fail after the provider expires the credential.

## Environment
- Service: production transcoder
- Provider: managed cloud transcription API

## Steps to Reproduce
1. Let a transcoder instance remain warm across metadata-token refresh cycles.
2. Receive a metadata token whose reported remaining lifetime is shorter than the fixed application cache duration.
3. Submit transcription work after the provider token has expired but before the application cache entry expires.

## Actual Result
The provider returns `401 ACCESS_TOKEN_EXPIRED`, which surfaces as a transcription HTTP 500 and leaves the video without a generated transcript.

## Expected Result
The transcoder should derive its cache lifetime from metadata `expires_in`, keep a safety window before expiry, invalidate a rejected token, and retry with a fresh credential.

## Regression?
Unknown.

## Evidence
Observed in current production logs. Identifying video values and complete provider payloads are intentionally omitted.

rabble:
## Summary

- Cache GCP access tokens using the metadata server's reported `expires_in`, capped by the existing 50-minute upper bound and shortened by a 5-minute skew.
- Treat provider `401 ACCESS_TOKEN_EXPIRED` responses as retryable and invalidate the rejected cached token for both Gemini and Google STT V2.
- Keep missing metadata `expires_in` graceful by falling back to the cache cap instead of failing token parsing.

## Reviewer takeover (dcadenas)

Pushed on top of the original commit; the summary above describes the first
commit only.

- Redact the bearer token from `MetadataAccessToken`'s `Debug` output.
- Invalidate a rejected token on the Cloud Tasks status-enqueue path too, and
  refresh the credential between its retries — it reads the same process-global
  cache.
- Route Gemini and Google STT V2 through one tested `gcp_provider_response_outcome`
  instead of two duplicated invalidate-then-classify blocks.

## Motivation

- A fixed 50-minute application cache could outlive the token returned by the metadata server, causing current transcription work to fail with provider 401 responses surfaced as HTTP 500s.

## Related Issue

- Closes #194

## Validation

- [x] `cargo test --manifest-path cloud-run-transcoder/Cargo.toml --locked`
- [ ] I could not run some validation locally, and I explained why below

## Manual Test Plan / Notes

- `cargo test --manifest-path cloud-run-transcoder/Cargo.toml --locked` — 151 passed (145 at the original head, plus 6 from reviewer takeover commits).
- `cargo check --tests --locked` passed with existing dead-code warnings.
- `cargo clippy --locked --all-targets --all-features` passed with existing clippy warnings outside this PR's touched code.

## Visuals / API Examples

- [x] No visual change

## Public Artifact Review

### DIFF
--- a/cloud-run-transcoder/src/main.rs
+++ b/cloud-run-transcoder/src/main.rs
@@ -4098,6 +4098,53 @@ mod tests {
         (url, request_count, handle)
     }
 
+    /// Like `spawn_status_server_with_bodies`, but records the `Authorization`
+    /// header of each request so a test can assert which credential was sent.
+    fn spawn_status_server_recording_auth(
+        responses: Vec<(u16, String)>,
+    ) -> (String, Arc<Mutex<Vec<String>>>, thread::JoinHandle<()>) {
+        let listener = TcpListener::bind("127.0.0.1:0").expect("bind local test server");
+        let url = format!("http://{}", listener.local_addr().expect("local addr"));
+        let bearers = Arc::new(Mutex::new(Vec::new()));
+        let seen = Arc::clone(&bearers);
+        let handle = thread::spawn(move || {
+            for (status, body) in responses {
+                let (mut stream, _) = listener.accept().expect("accept test request");
+                let _ = stream.set_read_timeout(Some(Duration::from_secs(2)));
+                let mut buf = [0_u8; 8192];
+                let read = stream.read(&mut buf).unwrap_or(0);
+                let request = String::from_utf8_lossy(&buf[..read]).to_string();
+                if let Some(header) = request
+                    .lines()
+                    .find(|line| line.to_ascii_lowercase().starts_with("authorization:"))
+                {
+                    seen.lock()
+                        .expect("bearer log")
+                        .push(header["authorization:".len()..].trim().to_string());
+                }
+                let reason = match status {
+                    200 => "OK",
+                    401 => "Unauthorized",
+                    409 => "Conflict",
+                    500 => "Internal Server Error",
+                    502 => "Bad Gateway",
+                    _ => "Status",
+                };
+                let response = format!(
+                    "HTTP/1.1 {} {}\r\nContent-Length: {}\r\nConnection: close\r\n\r\n{}",
+                    status,
+                    reason,
+                    body.len(),
+                    body
+                );
+                stream
+                    .write_all(response.as_bytes())
+                    .expect("write test response");
+            }
+        });
+        (url, bearers, handle)
+    }
+
     #[test]
     fn transcribe_audio_size_guard_rejects_empty_and_oversize() {
         assert!(ensure_transcribe_audio_size(0).is_err());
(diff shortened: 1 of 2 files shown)


rabble:
Production validation after deploying the green PR head:

- Cloud Build `104f2303-5e80-459f-b371-9c4bf35e7983` succeeded.
- Cloud Run revision `divine-transcoder-00050-fw4` is ready on image tag `98751ee` and serves 100% of untagged traffic.
- `/health` reports healthy.
- Initial live workload: 2 transcription starts; 1 uploaded a new transcript and 1 correctly skipped as already processing.
- No transcription errors, provider retries, or `ACCESS_TOKEN_EXPIRED` markers on the new revision.
- Deployment preserved queued status delivery and the existing cloud location.

NotThatKindOfDrLiz:
@rabble I took over the PR and narrowed it back to the token-expiry fix for #194.

What changed:
- Rebased onto current `origin/main` and rewrote the branch to one focused commit.
- Removed the deploy-guard changes from this PR so they can come back separately with complete env coverage and tests.
- Moved expired-token invalidation into shared GCP-token handling and wired it through both Gemini and Google STT V2.
- Updated metadata token parsing so missing `expires_in` falls back to the cache cap instead of failing all metadata attempts.
- Updated the PR body to match the actual diff and validation.

Validation:
- `cargo test --manifest-path cloud-run-transcoder/Cargo.toml --locked` — 145 passed.
- `cargo check --tests --locked` — passed with existing dead-code warnings.
- `cargo clippy --locked --all-targets --all-features` — passed with existing warnings outside the touched code.
- GitHub CI `test`, semantic PR title, and CLA are green; deploy jobs skipped.

Please review the takeover commit and merge when you're ready.

NotThatKindOfDrLiz (approved):
Approved after takeover.

The PR is now scoped to the #194 token-expiry fix only. The deploy-guard changes were removed from this branch, expired-token invalidation now covers both Gemini and Google STT V2, and metadata responses without `expires_in` degrade to the cache cap.

Validation passed locally and in CI:
- GitHub `test`, semantic PR title, and CLA

dcadenas (approved):
🔒 Cached token lifetime now tracks what the metadata server actually reports — the core fix is right. I pushed a few commits on top.

## What I pushed

- The metadata token struct derived `Debug` over a live bearer token. Replaced it with a redacting impl so a stray `{:?}` cannot put a GCP credential in the logs.
- The Cloud Tasks status-enqueue path reads the same process-global token cache but never dropped a rejected token, and its three retries replayed it. It now invalidates on an expired-token rejection and re-reads the credential between attempts.
- Gemini and Google STT V2 each carried their own copy of the invalidate-then-classify block. Both now go through one shared handler with tests. The invalidation on the transcription path — the path the issue actually reports — had no coverage before.

## Validation

- Transcoder suite 145 → 151. The new tests fail when the code they cover is reverted.
- `cargo check --tests`, `cargo clippy --all-targets --all-features`, plus the blossom-core and upload suites all pass. No new clippy warnings.
- Also ran the two Python suites CI runs: 23 and 77 pass.

## Notes, not blocking

- A metadata response without `expires_in` still falls back to the 50-minute cap, which is the assumption behind the original incident. It is deliberate and tested, and the metadata server does return the field in practice, so I left it — worth a second thought if you disagree.
- The diff carries `cargo fmt` reflow of code the change does not otherwise touch. Harmless, but it widens the diff around a credential-handling change.

Issue #183, opened 2026-08-07
feat(ops): run the cost burn-rate monitor on a schedule with alert delivery


rabble:
## Problem

`scripts/burn_rate_monitor.py` exists and works, but only runs when someone runs it. A script a human has to invoke is monitoring, not alerting.

## Why it is needed and why brain does not cover it

`divine-brain` already tracks infra cost well — `src/billing/forecast.ts` pulls Fastly billing v3 month-to-date and GCP billing-export costs, projects end-of-month per vendor, runs anomaly detection, and posts a daily digest to Slack with DB-backed suppression.

What it cannot do is **cadence**. Everything there is daily, and invoice data lags further. A TV-ad-driven traffic spike arrives in 60–120 seconds. The burn monitor reads Fastly real-time stats, so it is the only instrument that can see a cost spike while it is happening rather than the next morning.

Its accuracy is corroborated: modeled GCS egress derived from `compute_beresp_body_bytes` came out at $11.78/day against the actual GCP egress SKU at $11.45/day — 3% agreement between independent sources.

## Proposed shape

Cloud Scheduler → Cloud Run job → Slack webhook on non-zero exit. The script already supports this: `--webhook-url` (defaults to `$BURN_RATE_WEBHOOK_URL`), escalate-immediately / cooldown-on-steady / notify-on-recovery, and it alerts when collection *fails* so a blind monitor is never mistaken for a quiet one.

## Before Aug 18

- [ ] Deployed somewhere it runs unattended
- [ ] Alerts land in a channel people actually watch
- [ ] Thresholds reviewed — current defaults are calibrated against ~1,200 DAU and will need raising
- [ ] Rates in `RATES` replaced with contract rates once known (currently list prices, labelled as assumptions)

Issue #182, opened 2026-08-07
fix(edge): transcode status writes rejected by the generation guard still return 200


rabble:
## Problem

`POST /admin/transcode-status` returns `200 {"success": true}` even when the write is discarded as stale by the generation guard in `src/metadata.rs:313` (`stale_generation`). The caller cannot tell a successful write from a silently dropped one.

## How this was found

Repairing 15 videos after the 2026-08-07 outage, a reset of `status: pending` returned `200` for all 15 — and changed nothing for the 11 that had a recorded terminal event. Those kept serving `422 {"retryable": false}`. The repair looked complete and was not.

Resending the same payload with an explicit high `generation` worked immediately (422 → 202 → complete), confirming the guard was discarding the earlier writes.

## Why it matters

The generation guard itself is correct — it is what keeps out-of-order status events from resurrecting stale state. The problem is purely that rejection is indistinguishable from success at the API boundary. Any future repair or backfill tooling will hit this, and will report success while doing nothing.

## Proposed fix

Distinguish the two outcomes. Either:
- return `409 Conflict` (or `200` with `"applied": false` and the stored generation) when the write is dropped as stale, or
- have the response echo the generation actually stored, so a caller can detect a no-op

Either way the body should make it unambiguous.

## Acceptance criteria

- [ ] A stale-generation write is distinguishable from an applied write in the response
- [ ] Test covering applied / rejected-as-stale
- [ ] Repair tooling can detect a no-op without re-reading metadata

Issue #181, opened 2026-08-07
chore(infra): Cloud Run regional vCPU quota is at 90% before launch


rabble:
## Problem

Worst-case regional CPU allocation is now **627 of 700 vCPU (90%)** after raising the transcoder ceiling from 20 to 100 instances. Measured by `scripts/launch-capacity-check.sh`:

```
divine-transcoder         4 x 100 = 400 vCPU
blossom-upload-rust       1 x 100 = 100 vCPU
divine-blossom-upload  1000m x 100 = 100 vCPU
blossom-upload         1000m x  10 =  10 vCPU
docs-poc               1000m x  10 =  10 vCPU
cdn-view-subscriber    1000m x   3 =   3 vCPU
divine-connect-vip        1 x   2 =   2 vCPU
lookup-api                1 x   2 =   2 vCPU
                              total 627 / 700
```

Quota is enforced on CPU *actually* allocated at once, so this is a worst case rather than an admission check — but it leaves effectively no room to raise `maxScale` anywhere, and scale-out fails silently once the regional limit is hit.

## Two options, not mutually exclusive

**1. Request a quota increase.** Lead time is roughly a week, so it needs filing well before Aug 18 regardless of the final number. Free to ask.

**2. Reclaim ~110 vCPU by retiring superseded services.** `divine-blossom-upload` (100 vCPU) and `blossom-upload` (10 vCPU) appear superseded by `blossom-upload-rust`, which is what `cloud-run-upload/deploy.sh` targets and what the edge calls. **Needs confirmation from someone who knows their status** — I have not verified they are dead. Retiring them also removes services that could contend for CPU with the transcoder during a burst.

## Acceptance criteria

- [ ] Quota increase requested, or a decision recorded that current headroom is sufficient
- [ ] Status of `divine-blossom-upload` and `blossom-upload` confirmed; retired if superseded
- [ ] `scripts/launch-capacity-check.sh` reports headroom below the TIGHT threshold

Issue #180, opened 2026-08-07
fix(ci): deploy-process-blob fails Workload Identity Federation auth


rabble:
## Problem

The `deploy-process-blob` job has been failing on every merge to `main` for at least two merges (observed on #170 and #176):

```
##[error] google-github-actions/auth failed with: failed to generate Google Cloud
federated token for //iam.googleapis.com/***: {"error":"unauthorized_client"}
```

The GitHub Actions OIDC → GCP Workload Identity Federation trust is broken. Predates the current work.

## Impact today: none. Impact during launch week: real.

The last change to `cloud-functions/process-blob/` was `0c01698` on 2026-04-08, so nothing is currently sitting undeployed and the running function matches `main`.

But `process-blob` handles blob processing and the moderation webhook. If a fix is needed during the Aug 18 launch window, CI will fail silently-ish and it will have to go out via `scripts/deploy-cloud-function.sh` by hand — under time pressure, by someone who may not know the CI path is broken.

## Acceptance criteria

- [ ] WIF pool/provider/service-account binding repaired so `google-github-actions/auth` succeeds
- [ ] A merge to `main` shows `deploy-process-blob` green
- [ ] Until fixed, note the manual fallback in `docs/runbooks/deployment.md`

Worth doing before Aug 18 rather than discovering it during.

PR #196, opened 2026-08-11
docs(runbooks): document rollback traffic pinning and exported-env trap

rabble:
## Summary

Adds deployment traps from the 2026-08-07 transcoder outage to `docs/runbooks/deployment.md`:

1. **Exported shell variables override script defaults.** The hand-run deploy scripts resolve settings as `VAR="${VAR:-production-default}"`, so an exported `GCS_BUCKET=divine-blossom-media-staging` silently pointed the production transcoder at the staging bucket (4.5h of failed transcodes). There is no automated backstop for this today, and the runbook says so rather than implying one. It adds a shell check that derives the affected variable names from the deploy scripts themselves and reports which of them are exported, without printing any value.
2. **The two Cloud Run configuration checks read different places, on purpose.** The pre-deploy check reads the service's `spec.template`, because that is the baseline the next `gcloud run deploy` merges onto. The deploy-verification command reads the revision currently serving traffic, because that is what production is answering with. The two coincide normally and diverge after a rollback, so the runbook names which question each one answers.
3. **A traffic rollback pins the service.** `update-traffic --to-revisions` stops migration to latest, so a later deploy reports success while the new revision serves 0% traffic. Documents `--to-latest` as the required follow-up and a traffic-check command for the pinned window.

## Motivation

These traps surfaced during the Aug 7 incident response and were not yet documented in the deployment runbook.

## Related Issue

- Context: #179 and #182, both of which bit the incident response described here.
- Related follow-ups: #180, #181, #183.

## Validation

Docs-only change. No visual or behavioral surface.

- [x] `cargo test --manifest-path cloud-run-upload/Cargo.toml --locked`

## Manual Test Plan / Notes

Fact-checked against the tree:

- the `GCS_BUCKET` default in `cloud-run-transcoder/deploy.sh`
- that `gcloud run deploy` creates or updates the *service* and `--update-env-vars` merges onto its `spec.template`, which is what makes the template the right pre-deploy baseline and the serving revision the right deploy-verification baseline
- that `scripts/deploy-cloud-function.sh` carries the same exported-variable trap under `GCS_BUCKET_NAME` and `GCS_REGION`, so the shell check covers it too

Every embedded snippet was executed rather than only read:

- the `spec.template` reader and the serving-revision resolver, against synthetic Cloud Run v1 payloads covering a 100% revision, a pinned rollback, a tag-only traffic entry, and an empty traffic list
- the exported-variable check with a bucket variable and a secret variable exported — it names both and prints neither value, and it is silent under `env -i`
- `git diff --check`

Not tested: the snippets have not been run against a live Cloud Run service, only against payloads matching the v1 API shapes.

## Visuals / API Examples

- [x] No visual change

## Public Artifact Review

### DIFF
--- a/docs/runbooks/deployment.md
+++ b/docs/runbooks/deployment.md
@@ -68,14 +68,24 @@ GCP_PROJECT_ID=rich-compiler-479518-d2 ./scripts/deploy-cloud-function.sh
 Only `cloud-run-transcoder/deploy.sh` currently uses `--update-env-vars` and
 `--update-secrets`, so unnamed keys are preserved for transcoder deploys. Keys it
 *does* name are overwritten with the script's defaults, which may not match what
-is running. Compare before you deploy, and read names rather than values:
+is running. Compare against the revision serving traffic, not the service's
+`spec.template`, and read names rather than values:
 
 ```bash
-gcloud run services describe divine-transcoder \
+REVISION="$(gcloud run services describe divine-transcoder \
   --project rich-compiler-479518-d2 --region us-central1 --format=json \
   | python3 -c "
 import json,sys
-c=json.load(sys.stdin)['spec']['template']['spec']['containers'][0]
+traffic=json.load(sys.stdin).get('status',{}).get('traffic',[])
+serving=[t for t in traffic if t.get('percent')]
+print(max(serving,key=lambda t:t['percent']).get('revisionName','') if serving else '')
+")"
+
+gcloud run revisions describe "$REVISION" \
+  --project rich-compiler-479518-d2 --region us-central1 --format=json \
+  | python3 -c "
+import json,sys
+c=json.load(sys.stdin)['spec']['containers'][0]
 print('image:', c['image'])
 print('env:', sorted(e['name'] for e in c.get('env',[]) if 'value' in e))
 print('secrets:', sorted(e['name'] for e in c.get('env',[]) if 'valueFrom' in e))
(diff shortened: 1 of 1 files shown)


NotThatKindOfDrLiz:
@rabble I pushed takeover commit 4074789 to address the runbook contradictions from review.

What changed:
- the pre-deploy env inspection now reads the revision currently serving traffic instead of service spec.template
- the transcoder deploy verification now checks the serving revision's image, so pinned rollback traffic cannot look like a successful deploy
- the require-env-match paragraph now says this is post-#195 behavior, scopes the guard to a named subset, and documents the skip/fail-open cases including wrong PROJECT_ID lookup behavior
- the ambiguous rollback wording now matches #195's clearer explanation
- the PR body now reflects the dependency and validation accurately

Validation: git diff --check locally; GitHub test, semantic-pull-request, and license/cla are green. Please review the takeover changes and merge when ready.

NotThatKindOfDrLiz (approved):
Approved after takeover fixes.

I pushed 4074789 to make the runbook internally consistent: both pre-deploy config inspection and deploy verification now read the serving revision instead of spec.template, and the #195 guard paragraph is conditional/scoped to match the actual dependency and behavior. Checks are green.

dcadenas (approved):
✅ The three traps are the right ones to write down, and the incident detail makes them concrete enough to act on.

One thing needed fixing before this could land. The runbook promised a `scripts/require-env-match.sh` backstop arriving with #195, and no such script exists anywhere in the repository — #195 is a transcoder token-expiry fix touching two Rust files. Telling an operator that a deploy will abort on a mismatched bucket, in the same section that documents an outage caused by not checking manually, is worse than saying nothing.

### Changes I pushed

- Replaced the guard promise with the present-tense fact that nothing checks resolved variables today, and kept the design constraint for whoever adds one.
- Added a shell check for the trap that derives its variable names from the deploy scripts, so it cannot drift as the scripts grow. It covers `process-blob`, which carries the same trap under differently named variables, and it prints no values.
- Restored the pre-deploy configuration check to the service's `spec.template` and rewrote the closing paragraph to name which question each of the two checks answers.
- Narrowed the claim about which scripts resolve settings from the environment, and made the serving-revision block report plainly when nothing is serving traffic.

The `spec.template` change reverts the pre-deploy half of 4074789, so it is isolated in its own commit. `gcloud run deploy` creates or updates the *service*, and `--update-env-vars` merges onto its template — so the template, not the serving revision, is the baseline the next deploy merges onto. The serving revision stays correct for deploy verification, which asks what production is running rather than what the next deploy inherits. The two answers differ exactly during the pinned window this PR documents, which is why the runbook now states both.

### Checks

- `cargo check --tests --locked`, `cargo test -p blossom-core --locked`, `cargo test --manifest-path cloud-run-upload/Cargo.toml --locked`, `cargo test --manifest-path cloud-run-transcoder/Cargo.toml --locked`, `cargo clippy --locked --all-targets --all-features` — all pass
- `python3 -m unittest` for the transcoder and script suites — 23 and 71 tests pass. CI runs these as multi-command steps, so they were run by hand rather than through a single check command.
- Every embedded snippet executed against synthetic Cloud Run v1 payloads: a 100% revision, a pinned rollback, a tag-only traffic entry, and an empty traffic list.
- The exported-variable check run with a bucket variable and a secret variable exported: it names both, prints neither value, and is silent under `env -i`.
- Not covered: none of the snippets have been run against a live Cloud Run service.

### Follow-ups

- The env-match guard the runbook originally promised is still worth building, and nothing tracks it. An issue for it would be useful — the runbook now records what such a check has to compare and why parsing script defaults cannot work.

PR #199, opened 2026-08-13
feat(observability): durable structured logging for edge-proxied uploads

rabble:
## Summary

Adds one structured JSON upload-observability record for each edge request on:

- `PUT /upload`
- `POST /upload/init`
- `POST /upload/{id}/complete`

Records are sent to the configured Fastly Google Cloud Pub/Sub logging endpoint
and mirrored to stderr. Resumable chunk uploads are excluded because clients
send them directly to the upload service.

## Motivation

Origin logs cannot fully describe failures seen by the edge. This adds an
independent edge record, including send errors, response status, request-time
durations, route, declared media metadata, and a correlation ID. A send error
means the edge received no complete origin response; it does not by itself prove
whether origin received or processed the request.

## Implementation

- Keeps record formatting and outcome classification in `blossom-core` so the
  logic runs in the repository's executed unit-test suite.
- Emits from a router wrapper so early handler returns are recorded.
- Forwards `X-Request-Id` to the upload service on the three proxied routes.
- Uses monotonic clocks for durations and request occurrence time for delayed
  sink delivery.
- Sanitizes and bounds free-text fields and does not read request bodies or
  authorization headers for logging.
- Documents sink recreation, key handling, schema migration, and query examples
  in `docs/runbooks/edge-upload-observability.md`.

## Follow-up

Issue #201 tracks upload behavior improvements separately. The Pub/Sub to
ClickHouse subscriber and origin access-log correlation are also separate work.

## Validation

- `cargo check --tests --locked`
- `cargo test -p blossom-core --locked`
- `cargo test --manifest-path cloud-run-upload/Cargo.toml --locked`
- `cargo test --manifest-path cloud-run-transcoder/Cargo.toml --locked`
- `cargo clippy --locked --all-targets --all-features`
- `cargo check --target wasm32-wasip1 --locked`
- Both repository Python unittest suites

No visual change.

Issue #47, opened 2026-03-25
feat: add persistent logging for Fastly Compute service


dcadenas:
## Problem

The Fastly Compute service only has real-time ephemeral logging via `log-tail` (stderr). There are no persistent, searchable logs. When production issues are reported, we can't investigate because the logs are gone. Cloud Run logs have limited retention too (days, not weeks).

## Context

A user reported upload failures for 2 days. Client logs showed HTTP 500, 503, and connection resets from the Fastly edge. Cloud Run logs showed no record of the requests arriving, meaning Fastly returned the errors before proxying. We couldn't determine the root cause because neither system had historical logs for the period in question.

## Proposed solution

1. Create a GCS bucket (`divine-blossom-logs`) with a 7-day lifecycle delete policy
2. Add a GCS logging endpoint to the Fastly service
3. Replace `eprintln!` calls in `src/main.rs` with `log_fastly` + `log` crate to persist application logs to the endpoint
4. Include: request method, path, status code, backend used, response time, error details
5. Deploy

Gives persistent, searchable application logs with 7-day retention at minimal cost (~500 MB/week for current traffic).

## Also discovered during investigation

The `cloud_run_upload` Fastly backend has default timeouts (`FirstByteTimeout: 15s`, `BetweenBytesTimeout: 10s`). These should be reviewed separately — 15 seconds may be too tight for uploads that hit Cloud Run cold starts.

## Related

- No logging endpoints currently configured on the Fastly service
- Fastly Compute runtime limit: 120 seconds

dcadenas:
Fresh production measurements for this, taken 2026-09-01 while draining the NIP-62 vanish backlog. The issue was written from after-the-fact incident reports; these are live numbers, and they reproduce the exact signature described above.

## The outer VCL service is returning 503s to real users, continuously

`fastly stats historical` for service `ML7R82HKfmTaqTpHExIDVN` (outer VCL, `Divine.Video's website`), one-hour window ending 21:06 UTC:

```
132,929 requests
     27 status_503
```

Per-minute, only the minutes that had any:

```
20:01 -> 2   20:05 -> 1   20:09 -> 2   20:10 -> 1
20:14 -> 1   20:21 -> 2   20:29 -> 1   20:37 -> 7
20:40 -> 2   20:41 -> 1   20:46 -> 1   21:00 -> 2
21:01 -> 4   21:04 -> 2   21:05 -> 1   21:06 -> 1
```

[...]

PR #200, opened 2026-08-13
feat(edge): add persistent Fastly diagnostics

dcadenas:
## Summary

Fastly failures across the outer VCL cache and inner Compute service need a shared correlation key and durable diagnostic records.
This change adds repository-side logging sources and an operator runbook without changing live Fastly configuration.

## What Changed

- Preserve or generate a request ID at the outer edge, sanitize it once with a 64-character cap, and use the same value in Compute request records. UUIDs and trace IDs are preserved verbatim across both services; longer caller IDs keep a sanitized 64-character prefix.
- Emit structured outer diagnostics for every Fastly-generated 5xx in `vcl_error` before the existing synthetic 503 response, and add fail-open, error-only structured Compute logging for final 5xx responses. Rust panics are moved to the same endpoint via `fastly::log::set_panic_endpoint` once it exists; before that they keep their default stderr output.
- Remove the unconditional `[BLOSSOM ROUTE]` stderr line deliberately: route visibility now exists only as the normalized route category in persisted 5xx records.
- Share the transcript, quality-variant, and hash-path routing predicates between the Compute router and the diagnostic route category (blossom-core) so the two cannot drift.
- Document the exact endpoint names, manual VCL snippet placement, privacy limits, and later activation, alert, and validation work.

## Motivation

This implements the repository code and documentation portion of #47.
Persistent endpoints, activation, alerts, and production validation remain separate operator work.

## Testing

- `cargo test -p blossom-core --locked` (170 passed, including tests merged from main)
- `cargo clippy --locked --all-targets --all-features` (completed with existing warnings)
- `cargo test --manifest-path cloud-run-upload/Cargo.toml --locked` (41 passed)
- `cargo test --manifest-path cloud-run-transcoder/Cargo.toml --locked` (151 passed)
- `python3 -m unittest discover -s cloud-run-transcoder/tests -p 'test_*.py'` (23 passed)
- `python3 -m unittest discover -s scripts/tests -p 'test_*.py'` (77 passed)
- `git diff --check`

The branch now includes current `main` (#199, #204) merged in with the conflicts resolved; repository validation pack `310ffb8` ran every runnable declared command successfully on the merged head.
It still reports `blossom-core` as an uncovered exact project root despite separately passing derived clippy/tests, plus non-runnable toolchain-install, Python, and semantic-PR steps.
The Python suites were run manually and passed.

`cargo fmt --check` still reports pre-existing formatting drift in untouched files (six sites in `src/main.rs` and others elsewhere, all predating this branch).
Exact generated-VCL compilation, endpoint delivery, retention/access, alert configuration, and production correlation were not tested because they require the later Fastly operations phase.

## Manual Validation

After the required endpoints exist, validate all three VCL snippets in a cloned outer-service version before activation.
Then validate each failure stage separately: confirm a controlled Compute 5xx preserves its supplied request ID verbatim up to 64 characters after sanitization (longer IDs keep the 64-character prefix), and confirm a controlled outer failure preserves its supplied request ID and original Fastly error reason without producing a matching Compute record or changing the client response.

No visual change.

Part of #47


NotThatKindOfDrLiz (changes requested):
Thanks for the diagnostics work. I think this needs another pass before merge because the current behavior can mislead operators during the exact Fastly/Compute failure modes issue #47 is meant to diagnose.

Blocking changes requested:

1. `src/main.rs` / `docs/runbooks/fastly-5xx.md`: Compute panics, traps, OOMs, or wall-clock timeouts can still produce no `compute-diagnostics` record because `request_log::emit` only runs after `handle_request` returns. The runbook currently says an outer VCL 5xx with no matching Compute record means the request failed before Compute completed, which would misattribute guest traps. The pinned Fastly SDK has `fastly::log::set_panic_endpoint("compute-diagnostics")`; at minimum the runbook needs a caveat for non-returning Compute failures.

2. `blossom-core/src/request_diagnostics.rs` / `vcl/recv.vcl`: the 16-character request-ID cap means caller-provided UUIDs or trace IDs are not actually preserved across both services. Correlation between the two Fastly services works because both truncate the same way, but exact-match correlation against caller logs fails. Please either raise the cap enough for the IDs we expect to preserve, or update the PR/docs/validation language to say preserved prefix rather than preserved ID.

3. `vcl/error.vcl` / `docs/runbooks/fastly-5xx.md`: VCL diagnostics are logged only for `obj.status == 503`, while the triage text talks about outer VCL 5xx generally. Either widen the condition or make the runbook consistently 503-scoped.

Debt and accuracy fixes to include in the same pass:

- `blossom-core/src/request_diagnostics.rs`: `route_category` duplicates routing predicates and already drifts from them. Transcript casing and quality-variant matching should not be hand-maintained separately from the router/`QUALITY_VARIANTS`. The private `is_hash_path` also duplicates the exported blossom-core predicate with different semantics; if that difference is intentional, it needs a different name and rationale.
- `src/request_log.rs`: most `error_category` arms are unreachable because `emit` persists only 5xx records and only StorageError, MetadataError, and Internal currently map to 5xx. `backend_category` also uses a wildcard where future 5xx variants would silently log `backend: null`.
- `src/req_id.rs` and `README.md`: request-ID precedence docs are stale and omit `x-divine-edge-request-id`.
- `src/main.rs`: the old `[BLOSSOM ROUTE]` line is removed for non-5xx traffic. That may be the right privacy/volume choice, but it is an observable logging removal and should be stated deliberately.

I verified locally on this branch: `cargo check --tests --locked` succeeds with existing dead-code warnings, and `cargo test -p blossom-core --locked` passes 140 tests.

dcadenas:
@NotThatKindOfDrLiz thanks for the thorough review — all three blocking items and all four debt items are addressed in f8c1274, d49c82f, a4be9ad, and 6871a8d. The PR description's stale language (the "outer 503 diagnostics" wording and the uncaveated request-ID claim you called out in item 2) is corrected too.

**Blocking items**

- Panics/traps/OOM/timeouts no longer read as "never reached Compute". Rust panics are moved to `compute-diagnostics` via `fastly::log::set_panic_endpoint` (the pinned SDK's hook replaces the default stderr hook, so panic text stops appearing on stderr once the endpoint exists; until then the call fails open and stderr output is kept). The runbook's triage section now names panic/trap/OOM/wall-clock timeout as the alternative cause of a missing Compute record, and notes that panic records carry no request ID, so they correlate by UTC timestamp and POP instead.
- The request-ID cap is raised from 16 to 64 characters, in both the Compute sanitizer and the two VCL truncation sites. UUIDs (36), W3C trace ids (32), and `traceparent` values (55) now survive verbatim across both services; longer caller values keep a sanitized 64-character prefix, and the runbook, README, and PR description all say preserved prefix rather than preserved ID for that case.
- VCL diagnostics now log every Fastly-generated 5xx (`obj.status` 500–599) in `vcl_error`, matching the triage text. The client-facing synthetic JSON response stays 503-only, so other 5xx statuses keep Fastly's default error delivery and client behavior is unchanged.

**Debt items**

- `route_category` no longer hand-maintains its own predicates: the transcript parsers, `QUALITY_VARIANTS`, and the quality-variant predicates moved to blossom-core, and the category function now consumes those shared predicates in the router's arm order. The private `is_hash_path` duplicate is deleted in favor of the exported predicate; the behavioral consequence is that a 64-hex hash followed by junk segments now categories as "other" instead of "blob", which matches what the router actually does (404). A regression test pins the previously drifted cases (transcript casing, quality-variant shape, arm ordering).
- `request_log` now derives both the error category and the backend label from one exhaustive no-wildcard mapping over the error enum, with a test asserting the mapping exists for exactly the variants whose status codes are 5xx. A variant that newly maps to 5xx fails compilation, and a remapped existing variant fails that test, instead of silently logging `backend: null`.
- The request-ID precedence docs in `req_id.rs` and the README now list `x-divine-edge-request-id` first, and state that the header is trustworthy only once the outer VCL service is activated (until then a direct caller can set it and pin the logged ID).
- The `[BLOSSOM ROUTE]` removal is now stated deliberately in the runbook: it was unconditional per-request logging of method, full path, and host to the ephemeral Compute log, and route visibility now exists only as the normalized route category in persisted 5xx records. Operators who used it in `log-tail` for live debugging no longer have it at any status.

**Validation**

`cargo check --tests --locked`, `cargo test -p blossom-core --locked` (146 passed), `cargo clippy --locked --all-targets --all-features`, both Cloud Run Rust suites (41 and 151 passed), both Python suites run by hand (23 and 77 passed), and `git diff --check` are all green.

Coverage gaps from the repo's own declared check set: the validator reports `blossom-core` as an uncovered exact project root (its derived clippy and tests were run separately and pass), and the toolchain-install, Python, and semantic-PR CI steps are not locally runnable. `cargo fmt --check` still reports pre-existing drift in files this branch does not touch. Generated-VCL compilation, endpoint delivery, retention/access, alerts, and production correlation remain operator validation per the runbook.

NotThatKindOfDrLiz:
@dcadenas I pushed takeover commit 273b40e (`fix(edge): preserve diagnostic fallback visibility`) to address the two remaining diagnostics issues from re-review.

What changed:
- `src/request_log.rs` now mirrors serialized `divine.blossom.compute_request.v1` 5xx records to stderr even when the `compute-diagnostics` endpoint is absent, matching the redundancy pattern from `edge_upload_logs`.
- `vcl/recv.vcl` now uses explicit empty-string checks for request IDs, generates a fallback when the sanitized ID is empty, and writes that fallback to both `X-Request-Id` and `X-Divine-Edge-Request-Id` so VCL and Compute stay joinable.
- `docs/runbooks/fastly-5xx.md` now documents and requires validation of the all-filtered request-ID fallback case.

Validation I ran locally:
- `cargo test -p blossom-core --locked` — 170 passed
- `cargo check --tests --locked` — passed with existing dead-code warnings
- `cargo clippy --locked --all-targets --all-features` — completed with existing warnings
- `git diff --check` — clean

GitHub checks on the pushed head are green (`test`, `semantic-pull-request`, CLA; deploy/docker jobs skipped on PR as expected). Please review the takeover commit and merge when you are ready.

NotThatKindOfDrLiz (approved):
Approved after takeover commit 273b40e. The remaining diagnostics blockers are addressed: Compute 5xx records keep an stderr mirror when the durable endpoint is absent, and the VCL request-ID fallback now handles sanitized-empty values with one shared fallback ID. Local validation and current GitHub checks are green. Generated VCL compilation and endpoint delivery remain the operator validation called out in the runbook.

PR #205, opened 2026-08-16
fix(cache): align pending blob GET cache policy

### LINKED ISSUE #206: Track moderation-safe pending blob GET cache policy
Pending blob GET responses were using a different cache policy than the other public media routes, which limited edge offload while moderation was still in flight.

This issue tracks making GET cache behavior explicit for moderation states:

- Pending media may be public and edge-cacheable while moderation runs.
- Pending media must not use immutable one-year browser caching because browser caches cannot be purged after a later moderation status change.
- Active media can keep the immutable content-addressed blob policy.
- Restricted, age-restricted, banned, and deleted statuses must not be publicly cacheable.

Validation should include the Fastly edge compile/lint checks plus cache-header coverage for the status split.

rabble:
## Summary

- make GET and HEAD cache behavior explicit for each blob moderation status
- allow Pending public media to use public, purgeable edge caching with a 1-day browser TTL
- keep immutable 1-year browser caching reserved for Active media
- preserve private/no-store caching for restricted, age-restricted, banned, deleted, and admin responses
- apply the same status-driven policy to blob HEAD, thumbnail GET/HEAD, and audio GET/HEAD (previously immutable for Pending), and keep admin-bypass thumbnail and HLS content responses private
- move the status-to-policy split into `blossom-core` (`blob_cache_policy` / `cache_headers_for_policy`) so the exhaustive match and the header literals are covered by tests that execute in CI; unknown status fails closed to private

## Motivation

A production cache-policy mismatch made GET responses for Pending public media private/no-store even though HEAD and derivative routes treated the same media as publicly cacheable. This prevented effective origin offload during elevated traffic.

Pending media is still awaiting moderation, so the GET path should not give browsers an immutable one-year copy that cannot be purged after a later moderation status change.

Closes #206.

## Validation

- `cargo test -p blossom-core --locked` (175 passed)
- `cargo clippy --locked --all-targets --all-features` (exits 0; existing warnings remain)
- `cargo test --manifest-path cloud-run-upload/Cargo.toml --locked` (41 passed)
- `python3 -m unittest discover -s cloud-run-transcoder/tests -p 'test_*.py'` (23 OK)
- `python3 -m unittest discover -s scripts/tests -p 'test_*.py'` (93 OK)
- `git diff --check`

Note: `cargo test blob_get --locked` is not a usable local check for the Fastly binary; it fails at native link time on unresolved Fastly host symbols such as `_header_insert` and `_status_get`. Full `cargo fmt --check` also reports pre-existing formatting drift outside this PR, so I did not apply repo-wide formatting churn.

## Manual validation plan

1. Confirm a Pending public media GET returns `Cache-Control: public, max-age=86400`, `Surrogate-Control: max-age=31536000`, and the blob surrogate key.
2. Repeat the Pending GET and confirm a cache hit/offload in Fastly diagnostics.
3. Confirm an Active media GET keeps `Cache-Control: public, max-age=31536000, immutable`.
4. Change a test object to a restricted status and confirm the purge plus private/no-store response.
5. Monitor origin offload and error rate during rollout.

No visual change.

### DIFF
--- a/blossom-core/src/cache_policy.rs
+++ b/blossom-core/src/cache_policy.rs
@@ -71,11 +71,23 @@ pub fn status_requires_private_response(status: BlobStatus) -> bool {
     blob_cache_policy(status) == BlobCachePolicy::PrivateNoStore
 }
 
+/// The headers a serving route must send for a cache policy. Production routes
+/// apply this mapping instead of re-stating it, so the policy-to-headers
+/// decision is covered by the executing blossom-core tests.
+pub fn cache_headers_for_policy<'a>(policy: BlobCachePolicy, hash: &'a str) -> CacheHeaders<'a> {
+    match policy {
+        BlobCachePolicy::ImmutablePublic => immutable_blob_cache_headers(hash),
+        BlobCachePolicy::RevocablePublic => mutable_derivative_cache_headers(hash),
+        BlobCachePolicy::PrivateNoStore => private_no_store_cache_headers(hash),
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::{
-        blob_cache_policy, immutable_blob_cache_headers, mutable_derivative_cache_headers,
-        private_no_store_cache_headers, status_requires_private_response, BlobCachePolicy,
+        blob_cache_policy, cache_headers_for_policy, immutable_blob_cache_headers,
+        mutable_derivative_cache_headers, private_no_store_cache_headers,
+        status_requires_private_response, BlobCachePolicy,
     };
     use crate::types::BlobStatus;
 
@@ -142,11 +154,7 @@ mod tests {
                 "no-store",
             ),
         ] {
-            let headers = match policy {
-                BlobCachePolicy::ImmutablePublic => immutable_blob_cache_headers("abc123"),
-                BlobCachePolicy::RevocablePublic => mutable_derivative_cache_headers("abc123"),
-                BlobCachePolicy::PrivateNoStore => private_no_store_cache_headers("abc123"),
-            };
+            let headers = cache_headers_for_policy(policy, "abc123");
 
             assert_eq!(headers.cache_control, cache_control);
             assert_eq!(headers.surrogate_control, surrogate_control);
--- a/blossom-core/src/cache_policy.rs
+++ b/blossom-core/src/cache_policy.rs
@@ -1,8 +1,6 @@
 // ABOUTME: Shared cache policy for immutable blobs and repairable derivatives.
 // ABOUTME: Keeps browser and edge cache lifetimes testable outside Fastly Compute.
 
-use crate::types::BlobStatus;
-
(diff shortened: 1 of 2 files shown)


NotThatKindOfDrLiz:
@rabble I pushed one takeover commit, `1279113` (`fix(cache): keep pending blob GETs revocable`).

What changed:
- `Pending` GET responses are now public and edge-cacheable, but use the non-immutable one-day browser TTL (`Cache-Control: public, max-age=86400`) so browser copies stay revocable while moderation is still in flight.
- `Active` GET responses keep the immutable one-year blob policy.
- Restricted, age-restricted, banned, deleted, and admin GET responses use private/no-store headers; the admin branch now goes through the shared private-cache helper.
- Removed the pass-through `blob_response_is_private` wrapper and the duplicate core test; added GET cache-header coverage for Active/Pending/moderated statuses.
- Created and linked #206, and updated the PR body/validation notes.

Validation:
- `cargo test -p blossom-core --locked` (170 passed)
- `cargo clippy --locked --all-targets --all-features` (exit 0; existing warnings remain)
- `git diff --check`
- GitHub checks on `1279113`: `test` passed, `semantic-pull-request / build` passed; docker/deploy jobs skipped.

Notes:
- I did not apply repo-wide formatting churn. Full `cargo fmt --check` reports pre-existing formatting drift outside this PR.
- `cargo test blob_get --locked` is not a usable native local check for the Fastly binary because it fails at link time on unresolved Fastly host symbols.

Please review the takeover commit and merge when you are comfortable with the policy choice. I requested @divinevideo/platform for the required repo review; @dcadenas also still has an individual review request.

NotThatKindOfDrLiz (approved):
Approved after takeover fix.

I pushed `1279113` to keep Pending GET responses public and edge-cacheable while avoiding immutable one-year browser caching before moderation completes. Active keeps the immutable policy; moderated/admin states stay private/no-store. Local verification and GitHub checks are documented in the handback comment.

dcadenas (approved):
✅ The Pending cache split is right, and now it holds on every route that can serve the media.

### Changes I pushed

- Extended the revocable Pending policy to blob HEAD, thumbnail GET/HEAD, and audio GET/HEAD, which still carried the immutable one-year headers.
- Kept admin-bypass thumbnail and HLS content responses private, matching blob GET and the other routes.
- Moved the status-to-policy split and the header literals into `blossom-core`, so the exhaustive match and the exact header strings run in CI instead of being compile-only.
- Made an unknown moderation status fail closed to `private, no-store`.

The fail-closed default has no reachable behavior change today: every caller either has metadata or 404s first. It is isolated in its own commit.

### Checks

- `cargo test -p blossom-core --locked` (175 passed)
- `cargo clippy --locked --all-targets --all-features` (no new warnings)
- `cargo test --manifest-path cloud-run-upload/Cargo.toml --locked` (41 passed)
- transcoder and scripts Python test suites

@rabble the merge decision and the rollout timing stay with you.

PR #207, opened 2026-08-17
docs(runbooks): correct the edge upload sink runbook to what shipped

rabble:
## Summary

The edge upload observability runbook describes a world that stopped existing
when the subscriber shipped. This corrects it against what is actually running.

## What was wrong

**1. It said the sink did not exist.** A section headed *"Sink status —
incomplete, read before querying"* told readers there was no subscriber and no
table, that messages accumulated unacked, and that **"if the subscriber is not
built within 7 days, the oldest data is lost"**. The subscriber shipped in
divine-funnelcake #1017, migration `000231` created the table, and the Pub/Sub
backlog now sits at 0. A runbook warning about a dead deadline is worse than no
runbook.

**2. Every worked query was unrunnable.** They filtered on `timestamp`; the
shipped column is `occurred_at`. The proposed schema in this file was never what
got created:

| proposed here | shipped |
|---|---|
| `occurred_at_ms UInt64` + materialized `timestamp` | `occurred_at DateTime64(3)` |
| `Bool` columns | `UInt8` (this database has no `Bool` columns) |
| `PARTITION BY toYYYYMM(...)` | none (only 4 of 344 tables partition) |
| `MergeTree` | `ReplacingMergeTree` (Pub/Sub is at-least-once) |

## What it says now

- the live pipeline, with the ~10 minute lag explained as Fastly's log batching
  rather than subscriber lag
- the delivery semantics that affect a reader's counts: at-least-once with
  dedup on merge, insert-then-ack, and schema-1 rows carrying approximate
  timestamps
- the table definition that exists
- an explicit note that queries written against the old proposal will not run
- where the pre-subscriber backlog was archived

## Validation

Both rewritten queries were executed against the production table before
committing, not just eyeballed:

- route × outcome — returns `direct_put ok 2384 (97.27%)`, `send_failed 40 (1.63%)`
- size-bucket failure rate — returns rows across all five buckets

## Note

I wrote the original section, including the schema that turned out not to match
what shipped. The mismatch arose because the DDL was adjusted during review to
follow database conventions and the runbook was never brought along.

### DIFF
--- a/docs/runbooks/edge-upload-observability.md
+++ b/docs/runbooks/edge-upload-observability.md
@@ -2,8 +2,8 @@
 
 Structured logging for upload requests that pass through `media.divine.video`.
 
-Pipeline: Fastly Compute (`fastly-blossom`) → Google Cloud Pub/Sub → *(subscriber
-not yet built)* → ClickHouse
+Pipeline: Fastly Compute (`fastly-blossom`) → Google Cloud Pub/Sub →
+`edge-upload-log-subscriber` (Cloud Run) → ClickHouse `nostr.edge_upload_logs`
 
 Upload 5xx responses are also recorded on the separate `compute-diagnostics`
 sink with a route category rather than the upload-specific fields; see the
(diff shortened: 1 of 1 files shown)


NotThatKindOfDrLiz:
@rabble I pushed `dcaeff7` to take over the clear runbook fixes from review.

What changed:
- Replaced the stale Pub/Sub pull/ack-deadline guidance with language that matches the live subscriber draining and acking messages.
- Replaced the invalid `numUndeliveredMessages` subscription-resource command with a verified Cloud Monitoring API query, and added the Cloud Run service/region to inspect when backlog rises.
- Changed the remaining "how many?" guidance from Pub/Sub to ClickHouse.
- Tightened the archive object/window wording and added a schema-1 verification query before treating the archive as redundant.
- Updated the worked ClickHouse count queries to use `FINAL` and `schema = 2` where the new delivery/timestamp caveats require it.
- Linked the `edge-upload-log-subscriber` README in Related.

Verification:
- `git diff --check`
- Exact patched Cloud Monitoring backlog command returned `0` for `edge-upload-logs-sub`.
- Refreshed PR checks are green for `semantic-pull-request / build` and `test`; deploy-related jobs are skipped for this docs-only change.

@dcadenas your review request is still pending. Please review the updated branch; @rabble, merge when you are comfortable with the takeover commit.

NotThatKindOfDrLiz (approved):
Approved after takeover commit `dcaeff7`.

I re-reviewed the full docs diff after the push. The stale verification path is now corrected, the backlog check is a verified Cloud Monitoring query, the worked ClickHouse queries match the documented `FINAL`/`schema = 2` caveats, and CI is green for the refreshed head. Deploy jobs are skipped as expected for this docs-only PR.

Issue #198, opened 2026-08-12
fix(edge): Range requests to the bare blob return a cached NoSuchKey body as 206


realmeylisdev:
## Summary

`GET https://media.divine.video/{sha256}` with a `Range` header returns a **249-byte S3 `NoSuchKey` XML error body**, served with status **206** and a fabricated `Content-Range` claiming the object is 249 bytes. A plain `GET` of the same URL returns the real object.

Any client that range-requests a bare blob gets XML where it expected video. This is what made classic Vine videos unplayable on iOS (divinevideo/divine-mobile#7171) — AVFoundation always range-requests, so it always received the error body and failed with `AVFoundationErrorDomain -11829 "Cannot Open"`.

## Reproduce

```
$ curl -sS -H "Range: bytes=0-99" \
    https://media.divine.video/c541caf6c36d1da4b8264a03266a94fdb83b2ed6b5f91d198ac9e9ed92ecc678
<Error><Code>NoSuchKey</Code><Message>The object key does not exist.</Message>
<Resource>/divine-media-delivery/c541caf6c36d1da4b8264a03266a94fdb83b2ed6b5f91d198ac9e9ed92ecc678</Resource>...
```

Response headers on that request:

```
HTTP/2 206
content-range: bytes 0-99/249      <-- object is actually 855345 bytes
content-type: video/mp4            <-- body is XML
x-content-length: 855345           <-- origin's own header disagrees with content-range
cache-control: private, no-store   <-- yet it is being served from cache
age: 521
x-cache-hits: 12, 0
```

Plain GET of the same URL: `200`, `855345` bytes, correct.

Reproduced on three unrelated blobs, all reporting `/249`:

| blob | full GET | Range `bytes=0-99` |
|---|---|---|
| `c541caf6…ecc678` | 855345 | `bytes 0-99/249` |
| `8d4105dc…e8a410` | 768323 | `bytes 0-99/249` |
| `c4621006…5126f6` | 852009 | `bytes 0-99/249` |

A suffix range underflows on top of that — `Range: bytes=-65536` returns `content-range: bytes 18446744073709486329-248/249`, i.e. `249 - 65536` wrapped in unsigned 64-bit.

## Two separate defects

**1. Fastly is serving a cached error body for the Range variant.** `age: 521` with 12 cache hits, on a response marked `private, no-store`. The 206 status and synthesised `Content-Range` make it look like valid partial content, so clients do not treat it as an error — they treat it as a 249-byte file.

[...]

PR #209, opened 2026-08-18
fix(media): reject mirror responses that disagree with the addressed size

rabble:
## Summary

A `2xx` from the Fastly Object Storage mirror was treated as proof that the mirror held the object. `download_blob_from_fos` accepted any `200`/`206` without looking at the body, and `handle_get_blob` then set `Content-Type` from KV metadata over whatever arrived.

This adds `check_mirror_response` in `blossom-core`: the mirror stores content-addressed data, so its declared length must agree with the size recorded for the hash — `Content-Length` for a full body, the `Content-Range` complete-length for a partial one. A mismatch is treated as a mirror miss and the request falls through to GCS.

## Motivation

Issue #198 recorded the failure on three separate blobs: a 249-byte `NoSuchKey` XML document served as `206 Partial Content` with `Content-Range: bytes 0-99/249` and `Content-Type: video/mp4`, against objects of 855345, 768323 and 852009 bytes. A client cannot distinguish that from a very short video. AVFoundation range-requests everything, so on iOS it always received the error body.

Refs #198. That issue also reported the origin ignoring `Range` on the bare blob — a separate defect, and a different cause (the `vcl_miss` Range strip on an object that was never cacheable), fixed by #205. This PR addresses only the error-body half.

## Design notes

- **Fails closed.** An undeclared length, an unsatisfied `bytes 0-99/*` range, or a caller with no metadata to check against all reject. The only cost of a false rejection is one fetch from the authoritative origin.
- **The metadata-less case is admin-only.** `handle_get_blob` 404s a non-admin request before the download when metadata is absent, so `expected_size: None` reaches the mirror only on the admin bypass path. Mirror offload for ordinary traffic is unaffected.
- **A rejection heals the mirror.** The GCS fallback that follows is digest-verified before write-back (`try_write_back_to_fos`), so a full-body request overwrites the bad object under the same key. A range request does not — `write_back_decision` rejects `had_range` — so healing needs one uncached full GET.

## Validation

Run locally on this branch:

- `cargo test -p blossom-core --locked` — 184 passed (175 before, 9 new)
- `cargo check --tests --locked` — clean
- `cargo clippy --locked --all-targets --all-features` — exits 0; warning count unchanged against `main` (31 → 31), none in the changed code

Not run locally, and why:

- `cargo test --manifest-path cloud-run-upload/Cargo.toml` and the transcoder/Python suites — untouched by this change; CI covers them.
- **The wiring in `src/storage.rs` and `src/main.rs` has no automated test.** The Fastly binary's tests fail at native link time on unresolved host symbols (`_header_insert`, `_status_get`), so nothing local or in CI executes that adapter. The decision logic it calls is fully covered in `blossom-core`; the header extraction and the early return are covered by review only. #175 (Viceroy in CI) is the standing fix for that gap.
- **I could not reproduce the original `/249` response.** It does not occur today on the blob named in #198. The mechanism is established by reading the deployed code, not by observing the failure — so this is a fix for a demonstrated code path, not a verified repro. If FOS has since changed its behaviour on missing keys, this becomes belt-and-braces rather than a live fix, and it is still worth having: the mirror is a copy, and an unverified copy of content-addressed data should not be served.

## Manual validation plan

1. Confirm a bare-blob GET and a `Range` GET on a mirrored object still return the object, with `x-cache` behaviour unchanged.
2. Confirm byte offload does not drop after rollout — a bug here shows up as every mirror read falling through to GCS.
3. Watch for `[FOS] miss ... reason=untrusted_response` in Compute logs. A steady stream of `length_mismatch` means real mirror corruption; a steady stream of `size_unknown` means the mirror is not declaring lengths and the check needs revisiting.

No visual change.

### DIFF
--- a/blossom-core/src/read_through.rs
+++ b/blossom-core/src/read_through.rs
@@ -68,6 +68,77 @@ impl WriteBackDecision {
     }
 }
 
+/// Whether a response from the mirror can be trusted to be the addressed object.
+///
+/// The mirror is a copy of content-addressed data, so its declared length must
+/// agree with the size recorded for the hash. An S3-compatible origin can answer
+/// a `Range` request for a missing key with an error document under a `2xx`
+/// status; the body is then XML, but the serving route stamps the stored MIME
+/// type over it and the client receives a video-labelled error page.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum MirrorResponseCheck {
+    /// Declared length agrees with the addressed size.
+    Trusted,
+    /// The mirror did not declare a usable length.
+    SizeUnknown,
+    /// No addressed size to compare against.
+    ExpectedSizeUnknown,
+    /// The mirror declared a length the addressed object does not have.
+    LengthMismatch,
+}
+
+impl MirrorResponseCheck {
+    pub fn is_trusted(&self) -> bool {
+        matches!(self, MirrorResponseCheck::Trusted)
+    }
+
+    /// Stable short label for logging.
+    pub fn reason(&self) -> &'static str {
+        match self {
+            MirrorResponseCheck::Trusted => "trusted",
+            MirrorResponseCheck::SizeUnknown => "size_unknown",
+            MirrorResponseCheck::ExpectedSizeUnknown => "expected_size_unknown",
+            MirrorResponseCheck::LengthMismatch => "length_mismatch",
+        }
+    }
+}
+
+/// Read the complete-length field out of a `Content-Range` header value.
+///
+/// `None` for the unsatisfied form (`bytes 0-99/*`) and for anything that does
+/// not end in a decimal length.
+pub fn parse_content_range_total(value: &str) -> Option<u64> {
+    value.rsplit_once('/')?.1.trim().parse().ok()
+}
+
+/// Check a mirror response against the size recorded for its hash.
+///
+/// Every non-`Trusted` outcome costs one fetch from the authoritative origin and
+/// nothing else, so this fails closed: an unverifiable response is not served.
+pub fn check_mirror_response(
+    status: u16,
+    content_length: Option<u64>,
+    content_range: Option<&str>,
+    expected_size: Option<u64>,
+) -> MirrorResponseCheck {
+    let expected = match expected_size {
+        Some(size) => size,
+        None => return MirrorResponseCheck::ExpectedSizeUnknown,
+    };
+
+    let declared = if status == 206 {
+        content_range.and_then(parse_content_range_total)
+    } else {
+        content_length
+    };
+
+    match declared {
+        None => MirrorResponseCheck::SizeUnknown,
+        Some(size) if size == expected => MirrorResponseCheck::Trusted,
+        Some(_) => MirrorResponseCheck::LengthMismatch,
+    }
+}
+
 /// Decide whether a just-fetched object may be copied into the mirror.
 ///
 /// Every guard is a hard requirement: only a complete, length-declared,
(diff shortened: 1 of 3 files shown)


rabble (commented):
Independent review (second model, incident-context fresh). **Reads mergeable to me** — two non-blocking notes below. Leaving the approval itself to the requested platform reviewers since this comment comes from the author's own gh identity.

Verified independently rather than taken from the PR body:

- `download_blob_read_through` has exactly one call site at this head (`handle_get_blob`), so no other route silently loses mirror offload by passing `expected_size: None`.
- The non-admin + no-metadata 404 lands before the download, so `ExpectedSizeUnknown` really is confined to the admin bypass, as the body claims.
- `BlobMetadata.size` is a required `u64` with no serde default — any metadata row that parses carries a size, so the check always has a real comparator when metadata exists. A row with a *wrong* size (migration-written rows are the ones to watch) fails closed to GCS: offload cost, not correctness. The rollout plan's byte-offload watch covers exactly this.
- The trust decision runs on headers before the body is consumed, so a rejected mirror response doesn't pay the body transfer.
- The demonstrated #198 shape (206 whose Content-Range total is the error document's length) is test-covered, and the healing story holds: range fetches don't write back, but the outer-VCL Range strip on cache miss turns public-content misses into full GETs, which do.

Non-blocking notes:

1. For a 206 the check validates the Content-Range *total* but not that the returned range matches the *requested* range. An origin returning the wrong range with the right total would pass; S3-compatible origins don't do this, the demonstrated bug is caught, and a client would fail loudly on the mismatch. Reasonable follow-up hardening, not this PR.
2. Cosmetic: `parse_content_range_total("bytes */855345")` returns `Some` (the 416 form, unreachable from the 206 branch), while the test name says "absent_when_unsatisfiable". The assertion is fine; the name slightly overpromises.

The src/ wiring being review-only is honestly disclosed and tracked by #175; given CI can't link the edge binary's tests, putting the decision logic in blossom-core was the right call.

NotThatKindOfDrLiz (approved):
The mirror-size validation is correctly scoped and fails closed to the authoritative GCS origin. The diagnosis and wiring check out, the new decision logic is well tested, and the remaining observations are non-blocking cleanup or operational notes. Approved.

NotThatKindOfDrLiz:
## Review: Mirror response length validation

The fix is sound and the diagnosis is independently confirmed. All tests pass and the change is well-scoped. Ready to approve.

### Findings

**1. Test name vs. assertion (cosmetic) — `blossom-core/src/read_through.rs:270`**

Inside `content_range_total_is_absent_when_unsatisfiable_or_malformed`:
```rust
assert_eq!(parse_content_range_total("bytes */855345"), Some(855_345))
```

The test's name promises to show absent cases, but this line asserts the opposite: the RFC 9110 unsatisfied-range form `bytes */{total}` *is* successfully parsed. This contradicts the function's doc comment (`None for the unsatisfied form`). 

Unreachable in production today (the mirror never returns 416), so no failure scenario—but a future reader trusting the name will misunderstand the coverage. Recommend moving this line into a separate "is parsed" test or renaming to reflect what it actually checks.

**2. Hotpath allocation — `src/storage.rs:475-476`**

```rust
let content_range = resp
    .get_header_str("Content-Range")
    .map(|value| value.to_string());
```

Then immediately:
```rust
check_mirror_response(
    // ...
    content_range.as_deref(),
)
```

One heap allocation per mirror hit. Since nothing mutably borrows `resp` in between, pass the header directly:
```rust
let content_range = resp.get_header_str("Content-Range");
check_mirror_response(
    // ...
    content_range,
)
```

The function already handles `Option<&str>` correctly.

**3. Stray double blank line — `blossom-core/src/read_through.rs:275-276`**

Before `fn absent_flag_is_disabled()`. `cargo fmt` removes it; CI has no fmt gate, so this will persist as-is through review and need cleanup later.

### Operational notes (not defects)

**4. Load-bearing metadata.size — `src/metadata.rs:460`**

The check now depends on `metadata.size` being accurate. `update_transcode_status_with_metadata` independently overwrites it from the transcoder's `new_size` field. All six `send_status_webhook` call sites in `cloud-run-transcoder/src/main.rs` currently pass `None`, so dormant—but if a future writer ever populates it with a value that disagrees with the actual bytes stored, mirror offload silently drops to zero, surfacing only as `[FOS] miss ... reason=untrusted_response check=length_mismatch` in logs.

Suggest adding a doc comment on `new_size` naming the field as the invariant's one soft spot, so the maintainer of the transcoder webhook knows what could break delivery.

**5. Multipart responses always bypass — unstated behavior**

A `206` with `Content-Type: multipart/byteranges` carries no `Content-Range` header, so the check returns `SizeUnknown` and falls back to GCS. Correct and safe (nothing in this stack issues multi-range requests), but not mentioned in the fail-closed documentation.

### Verdict

**Ready to merge.** Findings 1–3 are one-line cosmetic cleanups; 4–5 are notes for operational awareness and maintainability, not changes needed before merge.

The change correctly limits scope to mirror offload without affecting the main cacheable path (Range is stripped in VCL before reaching Compute). The only cost of a false rejection is one GCS fetch.

Approved pending the three optional cosmetic fixes.

NotThatKindOfDrLiz (approved):
The fix is validated and ready. See my comment for optional cosmetic cleanup suggestions and operational notes.

PR #211, opened 2026-08-18
fix(cache): harden edge cache safeguards

### LINKED ISSUE #210: fix(cache): align edge policy and deploy safeguards
## Problem

The follow-up review of #208 found that the outer VCL conflates browser and edge cache policy, current deployment guidance still encourages routine global invalidation, and newly added production verification scripts are not reproducibly executable. The public incident working log also contains operational material that does not belong in this public repository.

## Impact

The current state can reduce intended short-lived negative caching, make the edge cache cold again during manual deploys, and give maintainers misleading confidence in verification that CI does not execute.

## Acceptance criteria

- Preserve the intentional short edge TTL for not-found responses while preventing private responses from entering shared cache.
- Make global invalidation an explicit manual action and document targeted surrogate-key invalidation as the normal path.
- Remove or reproducibly package and execute the new Node verification tools.
- Replace redundant or stale assertions with CI-executed cache-contract coverage.
- Remove the public incident working log from the current tree and escalate historical exposure separately if needed.
- Obtain the required platform-team review before merge.

Refs #208.

NotThatKindOfDrLiz:
## Summary

- key the outer VCL private-response guard to `Surrogate-Control`, preserving the intentional 60-second edge cache for 404 responses
- replace commit-message-triggered global purges with an explicit manual workflow input and update current operator guidance to prefer targeted surrogate-key purges
- remove the unreproducible Node verification scripts, redundant cache-policy test, and public incident working log added by #208
- add CI-executed contract tests for VCL ordering, manual purge gating, and current deployment guidance
- tag edge-cached 404 responses with the blob `Surrogate-Key` so targeted purges can evict them (review follow-up)

The guard is intentionally narrower than the previous `Cache-Control` check: browser `no-store` can coexist with an explicit short edge-cache policy, while private responses carry `Surrogate-Control: no-store` and remain excluded from shared caching.

## Motivation

PR #208 correctly stopped routine CI cache invalidation, but its VCL guard also bypassed deliberate negative caching and the repository still instructed manual deployers to globally purge. Its added production scripts had no dependency manifest or CI execution path. This follow-up closes those gaps without changing application response behavior.

## Linked issue

Closes #210.
Refs #208, #223.

## Manual validation

- `python3 -m unittest discover -s scripts/tests -p "test_*.py"`
- `python3 -m unittest discover -s cloud-run-transcoder/tests -p "test_*.py"`
- `actionlint` passed with only the repository custom-runner label diagnostic suppressed
- regression test demonstrated red against the original `Cache-Control` guard and green against the corrected surrogate policy

## Rollout

`vcl/fetch.vcl` is a hand-managed outer Fastly snippet and is not deployed by `fastly compute publish`. After merge, a platform maintainer must apply the updated snippet to the Fastly service and verify the 404 and private-response cache behavior before treating the production fix as live.

No visual change.


rabble (commented):
Reviewed against the merge base (#208) rather than current `main`, since `main` has moved several times today.

## The core fix is right, and the bug it fixes is real

`error_response` sets `Cache-Control: no-store` **together with** `Surrogate-Control: max-age=60` for 404s (`src/main.rs:5944-5947`) — a deliberate split so browsers don't store a moderation 404 while the edge absorbs the retry traffic for a minute. The previous guard matched `no-store` in `Cache-Control` and `return(pass)`-ed, so every 404 went to origin. Keying the guard on `Surrogate-Control` fixes that without weakening the private path. The ordering assertions in the new contract test pin it properly.

I also checked that removing `public_cache_headers_carry_the_token_the_vcl_matches` doesn't leave the new guard unpinned on the origin side: `every_cache_policy_sends_the_headers_its_name_promises` covers `PrivateNoStore -> Surrogate-Control: no-store`, executes in CI, and pins exactly the literal the new guard matches. That's the correct test for the new design — the removed one pinned the old key. Good trade, not a coverage loss. The import removal is consistent too; `private_no_store_cache_headers` is still referenced by production code in the same file.

## 1. Merge conflict with current `main` — resolve carefully

`git merge-tree origin/main <this branch>` conflicts in `blossom-core/src/cache_policy.rs`. `main` has since gained `immutable_storage_cache_policy`, `ImmutableStorageCachePolicy` and the two TTL constants, and `src/storage.rs` imports and calls them (lines 5 and 469).

A resolution that takes this branch's version of the file wholesale drops those and the edge crate stops compiling. Worth calling out explicitly because the conflicting hunk is adjacent to the test this PR deletes, so "keep ours" is the tempting resolution.

## 2. Three private responses drop from two layers to one

These set `Cache-Control: private, no-store` and **no** `Surrogate-Control`:

- `src/admin.rs:818`
- `src/main.rs:2600`
- `src/main.rs:2714`

All three deliberately serve content regardless of moderation status, for moderator review. Under the old guard `vcl_fetch` caught them. Under the new guard it does not — they carry nothing for it to match.

**They are safe today**, and I verified why: all three are `/admin/api/blob/...` routes, and `vcl_recv` only sends `^/[0-9a-fA-F]{64}` to `lookup`, so these take the `pass` path and are never stored. No live exposure.

But the comment retained directly above the guard still reads "This is the only thing keeping non-public responses out of the shared edge cache," and for these three that is now false — the recv path regex is. The two-independent-layer property that the private-cache work was built around is down to one for any response that classifies itself only via `Cache-Control`.

Small fix, inside this PR's intent: have those three emit both headers. `add_private_cache_headers` already does exactly that but is private to `main.rs`, so it's either lifting that helper or setting both inline in the three places. Happy to push it if you'd like — say the word and I will, otherwise it's yours.

## 3. Removing the file does not remove it from history

The working log is gone from the tree, but this repository is public and the file remains reachable in git history, on the #208 pull request page, and in any existing clone or fork. It carries service and config-store identifiers and backend topology.

#210's acceptance criteria say to "escalate historical exposure separately if needed" — I couldn't find that escalation tracked anywhere. Worth opening before this merges, so the deletion isn't mistaken for closure.

## 4. Purging now requires publishing

The purge step lives inside the `deploy` job, and that job now also triggers on `workflow_dispatch`. So a manual purge necessarily runs `fastly compute publish` first and creates a new service version. That's probably acceptable, but it's a behaviour change the summary doesn't mention, and it means "purge the cache" is no longer separable from "deploy". A separate job gated on the same input would decouple them.

Minor: the `purge_cache` input declares `required: true` alongside `default: false`. Harmless — GitHub renders a checkbox either way — but `required` does nothing for a boolean with a default.

## Note on scope

The title says "harden edge cache safeguards", and the purge gating and contract tests do that. The guard change is a *narrowing* — correct, and well justified by the 404 case, but narrower is the opposite direction from the title. Worth a sentence in the description so the next person reading the blame doesn't have to reconstruct why.

Not an approval — I don't have approval authority here, and `@divinevideo/platform` approval is required before merge.

rabble:
Shared-branch handback after resolving the review findings:\n\n- e4bac36 merges current main without rewriting history and preserves the newer immutable-storage cache helpers in the only conflict.\n- 6e93a3f restores two-layer private caching for all three moderator media paths, adds regression coverage, and separates manual global purge from Fastly publishing. A workflow dispatch can now purge without creating a service version; normal pushes still publish without globally invalidating content-addressed media.\n- The removed public working material now has a private internal escalation for the history/fork decision; no sensitive contents were copied into public tracking.\n\nValidation: cargo check --tests --locked; 177 blossom-core tests; 97 script tests; cargo clippy --locked --all-targets --all-features; diff checks. All passed with existing warnings. The TDD regression checks failed on the original branch and pass on this head.\n\n@NotThatKindOfDrLiz, the branch is back with you. @dcadenas, please re-review the updated head; requesting the platform team again because the previous request is stale.

rabble:
Main advanced with #209, so I merged the new `origin/main` into this shared branch as `de85ec5` (no conflicts) and reran the full validation on the resulting head:

- `cargo test -p blossom-core --locked` (185 passed)
- `python3 -m unittest discover -s scripts/tests -p 'test_*.py'` (97 passed)\n- `cargo clippy --locked --all-targets --all-features`\n- `git diff --check origin/main...HEAD`\n\n@divinevideo/platform: please review the current head. This supersedes prior-head approvals.

NotThatKindOfDrLiz:
Author update on the current head (`2b8d097`):

- rebased the branch directly onto current `main` (`85aa8a1`) and preserved the immutable-storage and mirror-size work while resolving the cache-policy conflict;
- kept both browser and edge private-cache headers on all three moderator media paths and retained the separate manual purge job;
- made the cold-fill measurement record self-contained so deleting the incident log no longer leaves a broken link or missing rationale;
- corrected the outer-VCL guard comment and retargeted the unclassified-response TODO to follow-up #223;
- updated the PR description with the intentional guard narrowing and the required hand-managed VCL rollout step.

Local verification passed: 97 script tests, 185 blossom-core tests, 177 edge tests under the CI-pinned Viceroy release, `cargo check --tests --locked`, clippy (existing warnings only), 41 upload-service tests, 151 transcoder Rust tests, 23 transcoder Python tests, `actionlint`, and diff/reference checks. Current-head CI passed in 1m34s; semantic-title and CLA checks are also green.

@rabble and @dcadenas, please re-review the updated head. The `@divinevideo/platform` request is live; approval from that team is still required before merge.

rabble:
Reviewed current head `2b8d097` after the author's history rewrite. The private response headers, decoupled manual purge, corrected VCL explanation, and tracked fail-closed TODO all address the outstanding findings.

Fresh local validation on the exact head passed: `cargo check --tests --locked`, 185 blossom-core tests, 97 script tests, clippy, and `git diff --check origin/main...HEAD`. Human platform approval is still required before merge.

rabble:
@NotThatKindOfDrLiz please provide Platform review on exact current head 2b8d097. The repair addresses the earlier review findings, current CI is green, and the outer VCL deployment will be performed separately after merge by cloning active v18, changing only the named fetch snippet, validating, activating, and smoke-testing without a global purge.

rabble:
@NotThatKindOfDrLiz please review and approve exact current head 8086340. I reviewed the two new commits: the 404 Surrogate-Key derives the same 64-hex prefix accepted by vcl/recv, is attached only to hash-addressed 404s, and the admin comment correctly removes the stale global-purge instruction. Fresh cargo check --tests, the 5 edge-cache contract tests, diff check, and CI (including Viceroy) pass on this head.

dcadenas (approved):
✅ The guard rekey survives a full enumeration of every cacheable response class: only the intended 404 delta remains, and the private paths still carry `Surrogate-Control: no-store` on both halves of the pair.

### Changes I pushed

- Tag edge-cached 404s on hash-addressed paths with the blob `Surrogate-Key`. Without it, the newly live 60-second negative cache could not be evicted by the targeted `fastly purge --key <hash>` this PR promotes, and the VCL comment promising key-based purge would have been false. The key matches the one the 200 for the same URL carries, including uppercase-hash URLs.
- Drop the last stale `purge --all` instruction in the tree, on the `skip_purge` bulk-approve comment. It now points at per-hash surrogate-key purge.

### Checks

- `cargo test -p blossom-core --locked`: 185 passed
- `python3 -m unittest discover -s scripts/tests -p "test_*.py"`: 98 passed
- `python3 -m unittest discover -s cloud-run-transcoder/tests -p "test_*.py"`: 23 passed
- Not run locally: `./scripts/run-edge-tests.sh` needs a Viceroy binary I do not have installed. The CI `test` job runs it and was still running when this review was published.

### Follow-up

- Removing the incident working log does not remove it from git history or existing clones. That exposure predates this PR and needs separate tracking; leaving it to the maintainers to decide whether to escalate.

PR #213, opened 2026-08-18
fix(cache): authorize before reusing media bytes

rabble:
## Summary

- keep credentialed media requests on the Compute pass path so NIP-98/BUD-01 validation runs on every request
- cache only the immutable storage subrequest after access is allowed, with full-response-only admission, streaming/range support, and hash-based purge keys
- expose the internal storage cache state for operational verification
- let public hash paths fall through to Fastly's generated shield selection before lookup
- add an end-to-end age-gated cache-boundary and range test using ephemeral NIP-98 signers

## Motivation

Protected media should reuse immutable bytes without allowing a cache hit to bypass viewer authorization. Keying an outer shared cache only on the presence of `Authorization` is insufficient because forged headers and different authenticated identities would share the same entry. This keeps authorization in front and moves byte reuse behind the validated access decision.

Linked issue: none; production cache/shielding follow-up.

## Validation

- `cargo test --manifest-path blossom-core/Cargo.toml --locked` (178 passed)
- `cargo build --target wasm32-wasip1 --release --locked`
- `shellcheck scripts/test-authenticated-storage-cache.sh`
- public VCL smoke suite: 16 passed, 0 failed
- live protected-media boundary: anonymous 401; forged NIP-98 401; valid ephemeral NIP-98 200; authorized repeat storage-cache HIT; cached range 206; anonymous remains 401 after warm
- live cold public probe traversed the configured shield before the delivery POP

## Deployment / manual follow-up

The combined package is active on Compute service version 344 and the recv snippet is active on VCL service version 18. Monitor pass rate, byte hit ratio, 5xx, and storage-cache HITs after merge. The hand-managed VCL service must remain aligned with `vcl/recv.vcl`.

No visual change.

### DIFF
--- a/blossom-core/src/cache_policy.rs
+++ b/blossom-core/src/cache_policy.rs
@@ -10,6 +10,36 @@ pub struct CacheHeaders<'a> {
     pub surrogate_key: &'a str,
 }
 
+/// Full content-addressed objects may live in the Compute service's internal
+/// read-through cache for a year. Access control still runs before that cache
+/// is consulted; this policy applies only to the storage subrequest made after
+/// the viewer has been authorized.
+pub const IMMUTABLE_STORAGE_CACHE_TTL_SECS: u32 = 31_536_000;
+pub const IMMUTABLE_STORAGE_CACHE_SWR_SECS: u32 = 86_400;
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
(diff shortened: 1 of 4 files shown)


NotThatKindOfDrLiz:
Follow-up review findings posted. See conversation for details on four findings: range request strategy (blocking design question), X-Divine-Storage-Cache header reliability, 304 handling, and VCL invariant test coverage.

Issue #219, opened 2026-08-18
test(smoke): cover private moderation-status cache behavior end to end


NotThatKindOfDrLiz:
## Summary

Add a credentialed post-deploy smoke path for moderation statuses whose incorrect behavior could disclose content: Restricted, AgeRestricted, Banned, Deleted, and admin-bypass responses.

## Motivation

PR #214 added public Active/Pending cache-policy consistency checks, but intentionally could not exercise private or hidden statuses. Unit tests cover the policy mapping; no live end-to-end check currently verifies that deployed routing, authorization, and cache headers preserve the private/hidden behavior.

## Acceptance criteria

- Use synthetic, dedicated fixtures; do not record real creator identifiers or credentials.
- Verify the anonymous status and cache behavior for Restricted, AgeRestricted, Banned, and Deleted blobs.
- Verify authorized owner/admin behavior where applicable, including `Cache-Control: private, no-store`.
- Verify GET, HEAD, thumbnail, audio, and HLS routes where each route is supported.
- Keep credentials outside the repository and document safe fixture provisioning/rotation.
- Make the check runnable after deploy; automate it only in an environment with the required secret bindings.

Follow-up to #214.

PR #221, opened 2026-08-18
test(smoke): validate route cache TTLs

NotThatKindOfDrLiz:
## Summary

- validate the expected cache TTL independently for blob GET, blob HEAD, and thumbnail HEAD
- use a per-run temporary directory for captured headers
- clean temporary files reliably with an EXIT trap

## Motivation

PR #214 compared the broad cache-policy class across routes but checked the exact `max-age` only on GET. A HEAD or thumbnail response could therefore keep the same class while returning the wrong TTL. Its cleanup was also unreachable after the final explicit exits.

The credentialed private-status end-to-end coverage remains separate and is tracked in #219.

## Linked work

Follow-up to #214.
Refs #219.

## Validation

- `bash -n scripts/smoke-test-vcl.sh`
- `shellcheck scripts/smoke-test-vcl.sh`
- `git diff --check`
- `bash scripts/smoke-test-vcl.sh media.divine.video` — 21 passed, 0 failed against the public Active fixture

## Manual validation plan

After the next Compute deploy, run the script against the default Active fixture and a dedicated Pending fixture. Confirm GET, HEAD, and thumbnail all report the same class and the expected TTL.

No visual change.

### DIFF
--- a/scripts/smoke-test-vcl.sh
+++ b/scripts/smoke-test-vcl.sh
@@ -144,6 +144,19 @@ cc_of() {
     | grep -i '^cache-control:' | cut -d' ' -f2- | tr -d '\r'
 }
 
+check_ttl() {
+  # Bounded match: a plain substring grep would let max-age=86400 pass on
+  # max-age=864000, silently hiding the TTL regression this block exists to catch.
+  local name="$1"
+  local expected_ttl="$2"
+  local cache_control="$3"
+  if echo "$cache_control" | grep -qE "max-age=${expected_ttl}([^0-9]|$)"; then
+    echo "  PASS: $name"; PASS=$((PASS + 1))
+  else
+    echo "  FAIL: $name (expected 'max-age=$expected_ttl')"; FAIL=$((FAIL + 1))
+  fi
+}
+
 check_policy_ttl() {
   local route="$1"
   local policy="$2"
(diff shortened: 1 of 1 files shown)


rabble:
Reviewed current head `3f8c5d3`. The per-run temp directory and EXIT cleanup remove the stale shared `/tmp` files, and the three route-specific TTL assertions close the coverage gap without changing production behavior.

Fresh validation passed: `bash -n`, ShellCheck 0.11.0, `git diff --check`, and the live smoke run against `media.divine.video` (21 passed, 0 failed). Human platform approval is still required before merge.

dcadenas (approved):
✅ The per-route TTL assertions close the gap #214 left, and the per-run temp dir fixes the unreachable cleanup. I pushed one commit so the new TTL checks cannot pass on a superstring of the expected TTL.

### Changes I pushed

- Added `check_ttl` with a bounded match (`max-age=<ttl>` followed by a non-digit or end of line). Before it, `max-age=864000` would have passed the `max-age=86400` assertion, hiding the exact regression this block exists to catch. The shared `check()` matcher is unchanged for its other call sites.

### Checks

- `bash -n scripts/smoke-test-vcl.sh`
- `shellcheck` (style level, clean)
- `python3 -m unittest discover -s scripts/tests -p 'test_*.py'`: 93 tests passed
- `git diff --check`
- Matcher self-test: exact value, value followed by comma, superstring, prefix, empty, and trailing CR all behave as intended

CI's `test` job was still running when I submitted this review. No CI job exercises this script; it is a manual operator tool, so the live smoke run in the PR body (21 passed, 0 failed against the Active fixture) is the end-to-end evidence.

### Follow-ups

- Anonymous thumbnail GET-vs-HEAD parity is not covered by #219 as written (that issue tracks the credentialed private-status end-to-end run). If we want that parity tracked, it needs one sentence there or its own issue. Not blocking.

rabble:
@dcadenas The exact-head code review is good, but the CLA check is now pending because commit `a692568` uses an author identity GitHub cannot associate with your account. Please add that commit email to your GitHub account or rewrite the commit with your verified signing identity, then re-approve the resulting exact head. I will not bypass the CLA gate.

PR #229, opened 2026-08-18
fix(deploy): refuse production transcoder deploys with the wrong bucket

### LINKED ISSUE #228: Prevent production transcoder deploys from using a non-production bucket
## Problem

The transcoder deploy script resolves `GCS_BUCKET` from the caller environment. An exported non-production value can therefore silently override the production default and deploy the production Cloud Run service against the wrong bucket, causing derivative and transcription failures.

## Acceptance criteria

- Production `divine-transcoder` deploys fail before Cloud Build when the resolved bucket is not `divine-blossom-media`.
- Non-production service/project deploys remain available.
- The guard has automated regression coverage.
- The deployment runbook documents the enforced invariant and the remaining exported-variable risk.

No media identifiers or customer information should be included in logs, tests, or documentation.

rabble:
## Summary

- refuse production `divine-transcoder` deploys when the resolved GCS bucket is not the production media bucket
- fail before Cloud Build so a contaminated shell cannot publish or activate a bad revision
- add executable regression coverage for both the refusal and an allowed non-production service
- update the deployment runbook with the enforced invariant and remaining exported-variable risk

## Motivation

The deploy script accepts exported variables as overrides. A stale non-production bucket in an operator shell can therefore redirect the production transcoder and cause derivative/transcription failures. The guard enforces the production project/service/bucket relationship at the source.

Closes #228.

## Validation

- `python3 -m unittest discover -s tests -p 'test_*.py'` — 32 passed
- `bash -n deploy.sh`
- `shellcheck deploy.sh`
- `git diff --check`

## Manual validation plan

Run the deploy script with stubbed `gcloud` under the production project/service and a non-production bucket; confirm it exits before `gcloud builds submit`. Repeat with a non-production service and confirm the build/deploy commands remain reachable.

No visual change.

### DIFF
--- a/cloud-run-transcoder/deploy.sh
+++ b/cloud-run-transcoder/deploy.sh
@@ -42,6 +42,16 @@ IMAGE_TAG="${IMAGE_TAG:-$(git -C "${REPO_ROOT}" rev-parse --short HEAD 2>/dev/nu
 IMAGE="gcr.io/${PROJECT_ID}/${SERVICE_NAME}:${IMAGE_TAG}"
 
 GCS_BUCKET="${GCS_BUCKET:-divine-blossom-media}"
+PRODUCTION_PROJECT_ID="rich-compiler-479518-d2"
+PRODUCTION_SERVICE_NAME="divine-transcoder"
+PRODUCTION_GCS_BUCKET="divine-blossom-media"
+if [ "${PROJECT_ID}" = "${PRODUCTION_PROJECT_ID}" ] && \
+  [ "${SERVICE_NAME}" = "${PRODUCTION_SERVICE_NAME}" ] && \
+  [ "${GCS_BUCKET}" != "${PRODUCTION_GCS_BUCKET}" ]; then
+  echo "Refusing production deploy: GCS_BUCKET must be ${PRODUCTION_GCS_BUCKET} for ${PRODUCTION_SERVICE_NAME} in ${PRODUCTION_PROJECT_ID}." >&2
+  echo "Unset the exported GCS_BUCKET and retry." >&2
+  exit 1
+fi
 WEBHOOK_URL="${WEBHOOK_URL:-https://media.divine.video/admin/transcode-status}"
 TRANSCRIPT_WEBHOOK_URL="${TRANSCRIPT_WEBHOOK_URL:-https://media.divine.video/admin/transcript-status}"
 TRANSCRIPTION_PROVIDER="${TRANSCRIPTION_PROVIDER:-gemini}"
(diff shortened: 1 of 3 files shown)


rabble:
@NotThatKindOfDrLiz the incident guard is green and ready for exact-head review at `5399d0ea7aa03119a74eb4742910a6c70820d473`. Live revision 00052 remains healthy at 100% traffic with zero staging-bucket or Forbidden log records since activation. Recovery audit found all 37 valid affected videos have HLS + 720p present; the sole remaining source is audio-only and cannot produce video derivatives.

rabble (commented):
Second-model review (posted from the author's gh identity, hence a comment — approval belongs to the requested platform reviewers). I also retitled the PR at Rabble's request (`guard production transcoder bucket` read as guarding the bucket); the semantic-title check re-ran green on the new title.

**Reads mergeable.** Verified rather than assumed:

- Extracted the PR's `deploy.sh` + new test file into a scratch dir and ran them: both guard tests pass — refusal fires with no `gcloud builds submit` recorded, and a non-production service stays deployable. CI's full suite is green on top of that.
- Checked the one plausible bypass: the guard keys on `PROJECT_ID` while the script also has `GCP_PROJECT_ID`. Not a hole — every `gcloud builds/run` command uses `--project "${PROJECT_ID}"`; `GCP_PROJECT_ID` only feeds the container's env. The dangerous direction (prod service + wrong bucket) is covered.
- Placement matches the runbook's own requirement: after variable resolution, before Cloud Build. The two read-only gcloud calls that can precede the refusal are harmless.
- The runbook edit is accurate and honestly scoped ("other resolved deploy variables are still unchecked").

One non-blocking note: the reverse cross-wire is still possible — a *staging* deploy carrying production env values (e.g. exported `GCP_PROJECT_ID`/`GCS_BUCKET` pointing at prod while `PROJECT_ID` is staging) sails through, since the guard only fires on the production project+service pair. That's explicitly the residual risk the runbook admits, so it's disclosure-consistent — just worth remembering that this guard protects the production *service*, not the production *bucket*.

NotThatKindOfDrLiz (approved):
Verified the production bucket guard at head 5a455c1. The guard is correctly placed before Cloud Build, its regression coverage passes, the permissive test now uses a synthetic project, and the runbook accurately documents the narrow invariant and residual exported-variable risk. Local validation: 32 Python tests, bash -n, shellcheck, and git diff --check. GitHub test, semantic-title, and CLA checks are green.

NotThatKindOfDrLiz:
@rabble I pushed takeover commit `5a455c1` to the PR branch. It replaces the real production project ID in the allowed-path test with a synthetic staging ID and rewraps the touched runbook paragraph. I also repaired the PR body's literal `\n` formatting so the validation and manual-validation sections render normally.\n\nVerification on the pushed head: all 32 transcoder Python tests pass; `bash -n deploy.sh`, `shellcheck deploy.sh`, and `git diff --check` pass; GitHub's test, semantic-title, and CLA checks are green. I found no remaining blockers and approved the current head. Please review the takeover commit and merge when ready. @dcadenas and @divinevideo/platform, the head moved and is ready for review.

PR #232, opened 2026-08-18
docs(deploy): correct a false claim in the guard test comment

rabble:
## Summary

The comment added in #229 says `divine-blossom-media-staging` is a deliberately fictional bucket name.
That is false: the deployment runbook records this exact staging bucket value as the one exported during the incident.
This corrects the comment without making a claim about current infrastructure state.

## Motivation

The fixture reproduces the incident input rather than using an arbitrary value.
Calling it fictional could lead a later reader to replace the value and weaken the regression test's connection to the incident.
The test value and behavior are unchanged.

## Validation

- `python3 -m unittest discover -s tests -p 'test_*.py'` from `cloud-run-transcoder/`: 32 passed
- Comment only; no behavior or visual change

### DIFF
--- a/cloud-run-transcoder/tests/test_deploy_guard.py
+++ b/cloud-run-transcoder/tests/test_deploy_guard.py
@@ -54,9 +54,9 @@ class TranscoderDeployGuardTest(unittest.TestCase):
     def test_production_deploy_rejects_nonproduction_bucket_before_build(self) -> None:
         # Reproduces the deploy that redirected the production transcoder to a
         # non-production bucket and broke transcoding for four and a half hours.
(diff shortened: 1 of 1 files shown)


dcadenas (approved):
✅ The incident fixture now documents the durable fact.

### Changes I pushed

- Anchored the comment to the recorded incident instead of current cloud state.

### Checks

- Transcoder Python suite: 32 passed
- All runnable declared and path-derived checks passed
- Declaration gaps: the transcoder root had no runnable declared local command; toolchain setup, the Viceroy script, Python CI steps, and the reusable semantic workflow were not runnable through the validator. The relevant Python suite was run manually.

